PR #1, opened 2026-04-17
Add comprehensive test suite for server services

CoderCoco:
## Summary
This PR adds a complete test suite for the server-side services using Vitest and AWS SDK client mocks. The tests cover core functionality for ECS, EC2, CloudWatch Logs, Cost Explorer, file manager, and configuration services.

## Key Changes
- **EcsService.test.ts**: 423 lines of tests covering task management, status retrieval, task definition registration, and lifecycle operations (start/stop)
- **FileManagerService.test.ts**: 251 lines of tests for file manager deployment, status tracking, and task orchestration
- **ConfigService.test.ts**: 174 lines of tests for Terraform output parsing, caching, region resolution, and configuration persistence
- **CostService.test.ts**: 113 lines of tests for Fargate cost estimation and actual cost retrieval from AWS Cost Explorer
- **LogsService.test.ts**: 99 lines of tests for CloudWatch Logs retrieval and stream handling
- **Ec2Service.test.ts**: 81 lines of tests for ENI public/private IP resolution
- **vitest.config.ts**: New Vitest configuration with Node environment and proper TypeScript support
- **package.json**: Added `test` and `test:watch` scripts, plus Vitest and related dependencies
- **tsconfig.server.json**: Excluded test files from TypeScript compilation

## Notable Implementation Details
- Uses `aws-sdk-client-mock` for mocking AWS SDK v3 clients
- Comprehensive mocking of logger to avoid console output during tests
- Tests cover both happy paths and error scenarios
- Validates AWS API call parameters and response handling
- Tests include edge cases like missing configuration, API failures, and null returns
- All tests use Vitest's `describe`/`it` syntax with proper setup/teardown via `beforeEach`

https://claude.ai/code/session_01ChQpGknsSfrqG4KYRBUqyQ

### DIFF
--- a/app/src/server/services/ConfigService.test.ts
+++ b/app/src/server/services/ConfigService.test.ts
@@ -1,0 +1,174 @@
+import 'reflect-metadata';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+vi.mock('fs', () => ({
+  readFileSync: vi.fn(),
+  writeFileSync: vi.fn(),
+  existsSync: vi.fn(),
+}));
+
+vi.mock('../logger.js', () => ({
+  logger: {
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+  },
+}));
+
+import { readFileSync, writeFileSync, existsSync } from 'fs';
+import { ConfigService } from './ConfigService.js';
+
+const mockExists = existsSync as unknown as ReturnType<typeof vi.fn>;
+const mockRead = readFileSync as unknown as ReturnType<typeof vi.fn>;
+const mockWrite = writeFileSync as unknown as ReturnType<typeof vi.fn>;
+
+function makeState(outputs: Record<string, { value: unknown }>): string {
+  return JSON.stringify({ outputs });
+}
+
+describe('ConfigService', () => {
+  let service: ConfigService;
+
+  beforeEach(() => {
+    service = new ConfigService();
+  });
+
+  describe('getTfOutputs', () => {
+    it('returns null and warns when state file does not exist', () => {
+      mockExists.mockReturnValue(false);
+      expect(service.getTfOutputs()).toBeNull();
+    });
+
+    it('parses outputs and fills defaults for missing keys', () => {
+      mockExists.mockReturnValue(true);
+      mockRead.mockReturnValue(
+        makeState({
+          aws_region: { value: 'us-west-2' },
+          ecs_cluster_name: { value: 'my-cluster' },
+          game_names: { value: ['minecraft', 'factorio'] },
+        }),
+      );
+
+      const outputs = service.getTfOutputs();
+      expect(outputs).not.toBeNull();
+      expect(outputs!.aws_region).toBe('us-west-2');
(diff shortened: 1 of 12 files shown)


app/src/server/services/ConfigService.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    error: vi.fn(),
+  },
+}));
+
+import { readFileSync, writeFileSync, existsSync } from 'fs';
+import { ConfigService } from './ConfigService.js';
+
+const mockExists = existsSync as unknown as ReturnType<typeof vi.fn>;
+const mockRead = readFileSync as unknown as ReturnType<typeof vi.fn>;
+const mockWrite = writeFileSync as unknown as ReturnType<typeof vi.fn>;
```
CoderCoco: I don't like typing things as unknown ant then something else. Can you use a testing framework like jest?

app/src/server/services/ConfigService.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    it('uses aws_region from outputs when available', () => {
+      mockExists.mockReturnValue(true);
+      mockRead.mockReturnValue(makeState({ aws_region: { value: 'ap-south-1' } }));
+      expect(service.getRegion()).toBe('ap-south-1');
+    });
+
+    it('falls back to AWS_DEFAULT_REGION env var when outputs unavailable', () => {
+      mockExists.mockReturnValue(false);
+      const prev = process.env['AWS_DEFAULT_REGION'];
+      process.env['AWS_DEFAULT_REGION'] = 'eu-west-3';
```
CoderCoco: we should really wrap environment access into a service function or something. Directly modifying process.env can be dangerous in tests and could result in flakyness

CoderCoco (commented):
This is for claude but overall looks good, take a look at the comments.

Some other high level notes:

Tests must be written in the form of `it("should do something")` so that you can read them in a natural language. Additionally even though these are tests, we should be documenting any functions and variables with tsdoc comments so we understand what they are for later.

app/src/server/services/ConfigService.test.ts
CoderCoco: Good call — I've swapped the `as unknown as ReturnType<typeof vi.fn>` pattern for `vi.mocked(fn)`, which is vitest's typed helper for module mocks and gives proper autocomplete on `.mockReturnValue`, `.mock.calls`, etc. For service-shaped stubs (EcsService, Ec2Service, etc.) I replaced the double-cast with `Partial<T>` + a single `as T`.

I kept vitest rather than switching to jest — jest would need additional config for ESM in this project and the typing friction was really about the casting pattern, which `vi.mocked` solves cleanly. Happy to migrate to jest if you'd still prefer; just let me know.

---

app/src/server/services/ConfigService.test.ts
CoderCoco: Agreed. I extracted env access into `ConfigService.readEnvRegion()` so nothing in the service directly reads `process.env` at call time. Tests now stub it via `vi.spyOn(service, 'readEnvRegion').mockReturnValue(...)` — no more mutating `process.env` and no cross-test leakage.

---

PR #9, opened 2026-04-17
Refactor Discord command handling into pluggable SlashCommand classes

CoderCoco:
## Summary

Refactored the monolithic `DiscordBotService` command dispatcher into a registry-based architecture where each Discord command is its own `SlashCommand` class. This improves testability, maintainability, and makes it easier to add new commands without modifying the core service.

## Key Changes

- **Extracted command logic into dedicated classes:**
  - `ServerStartCommand`, `ServerStopCommand`, `ServerStatusCommand`, `ServerListCommand` each implement the `SlashCommand` interface
  - Commands handle their own permission checks, option parsing, and service dispatch
  - Shared autocomplete logic lives in `GameOptionSlashCommand` base class

- **Created `SlashCommandRegistry`:**
  - Central lookup for commands by name
  - Builds all command descriptors for Discord REST API in one place
  - Replaces inline `buildCommands()` method and command-name switching

- **Introduced `CommandInvoker`:**
  - Encapsulates the (guildId, userId, roleIds) tuple needed by every command
  - Centralizes role ID extraction from both `GuildMember` and `APIInteractionGuildMember` shapes
  - Exposes `canRun(game, action)` that defers to `DiscordConfigService`

- **Simplified `DiscordBotService`:**
  - Removed ~300 lines of command dispatch, permission checking, and formatting logic
  - Now a thin transport layer: validates guild/allowlist, looks up command in registry, delegates to `command.execute()` or `command.autocomplete()`
  - Dependency injection reduced from 3 services to 2 (registry + discord config)

- **Extracted `formatGameStatus()`:**
  - Stateless utility function for rendering game status
  - Reused by both single-game and multi-game status commands

## Implementation Details

- Commands are Nest `@Injectable()` providers wired in `discord.module.ts`
- `SlashCommand` abstract base defines `name`, `action`, `build()`, `execute()`, and `autocomplete()` contract
- `CommandContext` and `AutocompleteContext` interfaces provide type-safe parameter passing
- Permission order remains in `DiscordConfigService.canRun()` as documented
- Tests updated to use `makeBot()` helper that wires the full DI graph with real command classes

https://claude.ai/code/session_01KbMbYpy97LSnWHsCC7w4A2


app/src/server/services/DiscordBotService.ts
```diff
@@ -322,224 +282,35 @@ export class DiscordBotService {
...
-        command: interaction.commandName,
-        game,
-      });
-      await interaction.reply({
-        content: `You don't have permission to ${action} **${game}**.`,
-        flags: MessageFlags.Ephemeral,
-      });
+    if (isAutocomplete) {
+      const focused = interaction.options.getFocused(true);
+      await command.autocomplete({ interaction, invoker, focused: { name: focused.name, value: focused.value } });
```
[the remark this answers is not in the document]
CoderCoco: Declined — the autocomplete path can't realistically throw: `invalidateCache()` is a field assignment, `getTfOutputs()` has its own internal try/catch returning `null` on error, `canRun()` is pure, and `respond()` already has `.catch(() => undefined)`. Adding defensive error boundaries for scenarios that can't happen clutters the dispatcher. Same pattern as the pre-refactor `handleAutocomplete`.

---

app/src/server/services/DiscordBotService.ts
```diff
@@ -322,224 +282,35 @@ export class DiscordBotService {
...
-   */
-  private extractRoleIds(member: InteractionMember | null): string[] {
-    if (!member) return [];
-    if (member instanceof GuildMember) {
-      return [...member.roles.cache.keys()];
-    }
-    // APIInteractionGuildMember: `roles` is a plain array of role IDs.
-    return [...member.roles];
+    await command.execute({ interaction, invoker });
   }
```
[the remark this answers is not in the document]
CoderCoco: Declined — each command's `execute()` already wraps its ECS call in try/catch with an `editReply('❌ Command failed…')` fallback. The pre-guard `reply()` calls (e.g., "Game is required", "You don't have permission") sit outside that try/catch just like they did pre-refactor in `handleInteraction` — this PR preserves the exact same exception shape. Not a regression introduced here.

---

app/src/server/discord/SlashCommandRegistry.ts
```diff
@@ -0,0 +1,47 @@
...
+  private readonly commands: Map<string, SlashCommand>;
+
+  constructor(
+    start: ServerStartCommand,
+    stop: ServerStopCommand,
+    status: ServerStatusCommand,
+    list: ServerListCommand,
+  ) {
+    this.commands = new Map();
+    for (const cmd of [start, stop, status, list] as SlashCommand[]) {
```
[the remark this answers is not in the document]
CoderCoco: Declined — command names are hardcoded string literals on each class and the registry constructor takes the four commands by concrete type, so duplicate names can only happen via a deliberate copy-paste typo that tests would catch (the command's test asserts its `name`). Adding a runtime duplicate-detection throw guards against a scenario that can't occur without a code-review miss on both sides. Nitpick.

---

app/src/server/discord/commands/ServerListCommand.ts
CoderCoco, line 1: Ensure that you document all public fields and methods.

CoderCoco: Or at least use `/** @inheritDoc **/`

app/src/server/discord/SlashCommand.ts
```diff
@@ -0,0 +1,58 @@
...
+ *   filtering too: `/server-start <tab>` only shows games the user has the
+ *   `start` action on.
+ * - `autocomplete()` handles option-autocomplete; commands without options
+ *   inherit the default no-op.
+ */
+export abstract class SlashCommand {
+  /** The Discord command name (without the leading slash). */
+  abstract readonly name: string;
+  /** Permission bucket for {@link DiscordConfigService.canRun} lookups. */
+  abstract readonly action: DiscordAction;
```
CoderCoco: Would it make more sense to create a protected constructor that sets these variables?

```ts
protected constructor(readonly name: string, readonly action DiscordAction)
```

app/src/server/discord/SlashCommandRegistry.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+@Injectable()
+export class SlashCommandRegistry {
+  private readonly commands: Map<string, SlashCommand>;
+
+  constructor(
+    start: ServerStartCommand,
+    stop: ServerStopCommand,
+    status: ServerStatusCommand,
+    list: ServerListCommand,
```
CoderCoco: Since we are defining commands as a `SlashCommand` can we actually just pass in a single vairable here and have dependency injection resolve all types to that single array?

```ts
constructor(commads: SlashCommand[]) {
    // Existing code here
}
```

app/src/server/discord/SlashCommandRegistry.ts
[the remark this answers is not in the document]
CoderCoco: Nest doesn't support auto-injecting "all providers of type T" as an array the way Spring does with `List<T>` — a bare `constructor(commands: SlashCommand[])` won't resolve. Two realistic options if you want the registry to drop its knowledge of the concrete command classes:

**A. Factory provider with a custom token** — moves the four-element list from the registry constructor into `discord.module.ts`:
```ts
export const SLASH_COMMANDS = Symbol('SLASH_COMMANDS');
// in providers:
{ provide: SLASH_COMMANDS,
  useFactory: (...cmds: SlashCommand[]) => cmds,
  inject: [ServerStartCommand, ServerStopCommand, ServerStatusCommand, ServerListCommand] }
// registry:
constructor(@Inject(SLASH_COMMANDS) commands: SlashCommand[]) { ... }
```

**B. `DiscoveryService`** from `@nestjs/core` — enumerate providers at runtime and filter `instanceof SlashCommand`. More "automatic" but relies on reflection + an extra module import; overkill for four commands.

Both still require listing the four classes somewhere (A: module's `inject` array; B: module's `providers` array, which it already does). The net benefit is the registry no longer imports concrete command classes. Happy to apply A if you want it — or keep the current shape, since adding a command already touches the module anyway.

---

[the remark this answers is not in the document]
CoderCoco: Applied in e8a8f0e — `SlashCommand` now has a `protected constructor(public readonly name: string, public readonly action: DiscordAction)`, `GameOptionSlashCommand` forwards both to `super`, and each concrete command calls `super('server-start', 'start', ...)` etc. instead of redeclaring the fields. Tests still pass.

---

app/src/server/discord/commands/ServerListCommand.ts
CoderCoco: Done in e8a8f0e — added TSDoc on the `CommandContext`/`AutocompleteContext` fields, documented the new `SlashCommand` protected-constructor params, and tagged every overridden `build()` / `execute()` / `autocomplete()` across the four command classes with `/** @inheritDoc */`. `name`/`action` are now inherited from the base so there's nothing left to document at the subclass level for those.

---

CoderCoco: Applied in e8a8f0e — every overridden `build()` / `execute()` / `autocomplete()` in the four command classes and `GameOptionSlashCommand` now has a `/** @inheritDoc */` tag.

---

app/src/server/discord/SlashCommand.ts
CoderCoco: Applied in e8a8f0e — `SlashCommand` now has `protected constructor(public readonly name: string, public readonly action: DiscordAction)`, `GameOptionSlashCommand` forwards both to `super()`, and each concrete command calls `super('server-start', 'start', ...)` (etc.) from its own constructor instead of redeclaring the fields. Tests and build pass.

---

app/src/server/discord/SlashCommandRegistry.ts
CoderCoco: Replied in the Copilot thread above — short version: Nest can't auto-inject "all providers of type `SlashCommand`" as an array. The cleanest equivalent is a factory provider with a custom token (Option A there), which moves the four-element list from the registry constructor into `discord.module.ts`. The registry loses its import of the concrete command classes; the module still has to list them. Let me know if you want me to apply A.

---

CoderCoco: This is incorrect. Below is an example on how this could be done:

```ts
import { Injectable, Inject, Module } from '@nestjs/common';

// 1. Define the Abstract Class
@Injectable()
export abstract class Handler {
  abstract handle(): string;
}

// 2. Concrete Implementations
@Injectable()
export class HandlerA extends Handler {
  handle() { return 'A'; }
}

@Injectable()
export class HandlerB extends Handler {
  handle() { return 'B'; }
}

// 3. Register in Module
@Module({
  providers: [
    { provide: Handler, useClass: HandlerA, multi: true },
    { provide: Handler, useClass: HandlerB, multi: true },
  ],
})
export class HandlerModule {}

// 4. Inject the Array
@Injectable()
export class ConsumerService {
  constructor(
    @Inject(Handler) private readonly handlers: Handler[]
  ) {
    // handlers will be [HandlerA, HandlerB]
    console.log(this.handlers.map(h => h.handle())); // ["A", "B"]
  }
}
```

CoderCoco: This is Angular DI, not Nest. Nest never adopted the `multi: true` flag — if you register the same token twice in Nest, the later registration overwrites the first (same-token providers don't accumulate). `@Inject(Handler)` always resolves to a single `Handler`, not `Handler[]`.

Evidence from this project's installed `@nestjs/common@10.4.x`:

```ts
// node_modules/@nestjs/common/interfaces/modules/provider.interface.d.ts
export interface ClassProvider<T = any> {
  provide: InjectionToken;
  useClass: Type<T>;
  scope?: Scope;
  inject?: never;
  durable?: boolean;
  // no `multi` field
}
// Same for ValueProvider, FactoryProvider, ExistingProvider.
```

Happy to prove it with a throwaway vitest/Nest test in this repo if you'd like — the snippet you pasted will compile (TypeScript doesn't know `multi` isn't a real key on an object literal satisfying `ClassProvider`) but at runtime only the last-registered provider for token `Handler` survives and `this.handlers` is `HandlerB` (the instance, not an array). See also [nestjs/nest#1045](https://github.com/nestjs/nest/issues/1045) and [nest#4319](https://github.com/nestjs/nest/issues/4319) where Kamil explicitly declines multi-injection in favor of factory providers — which is the Option A I proposed above.

Option A with `useFactory` + a `SLASH_COMMANDS` token is the real Nest-idiomatic way to inject "all the commands" as a single array. Want me to apply it?

---

CoderCoco: Alright then can we map everything to a single token that way we have a single place where the array is mapped (in case this grows in the future).

This is an example I found from [this thread](https://github.com/nestjs/nest/issues/4786#issuecomment-755032258):

```ts
    {
      provide: 'Tasks',
      useFactory: (email, sms) => [email, sms],
      inject: [EmailTask , SmsTask ],
    },
```

```ts
constructor(@Inject('Tasks') private tasks: Array<ITask>) {
  }
```

app/src/server/discord/SlashCommandRegistry.ts
```diff
@@ -0,0 +1,47 @@
...
+    return this.commands.get(name);
+  }
+
+  /** Every registered command, in insertion order. */
+  all(): SlashCommand[] {
+    return [...this.commands.values()];
+  }
+
+  /** Serialize every command's descriptor for a Discord REST PUT. */
+  buildAll(): RESTPostAPIChatInputApplicationCommandsJSONBody[] {
```
CoderCoco: Where was this type added? This name is way too long.

app/src/server/discord/SlashCommandRegistry.ts
CoderCoco: Applied in 1f2f77a. The registry now takes `@Inject(SLASH_COMMANDS) commands: SlashCommand[]` and has no import of any concrete command class. The array is assembled in one place — `discord.module.ts`:

```ts
const slashCommandsProvider = {
  provide: SLASH_COMMANDS,
  useFactory: (...cmds: SlashCommand[]): SlashCommand[] => cmds,
  inject: [ServerStartCommand, ServerStopCommand, ServerStatusCommand, ServerListCommand],
};
```

Used a `Symbol('SLASH_COMMANDS')` for the token instead of a string — stronger uniqueness guarantee; same shape otherwise. Adding a new command is now a single-file edit to `discord.module.ts`: register the provider + add it to the factory's `inject` and returned array. Tests still pass.

---

app/src/server/discord/SlashCommandRegistry.ts
CoderCoco: Not one I added — it's discord.js's own type for a chat-input command REST payload (exported from `discord.js`). I agree the name is painful to read; aliased to `SlashCommandDescriptor` in 1f2f77a:

```ts
// SlashCommand.ts
export type SlashCommandDescriptor = RESTPostAPIChatInputApplicationCommandsJSONBody;
```

Now used as the return type of `SlashCommand.build()` and `SlashCommandRegistry.buildAll()`.

---

PR #116, opened 2026-05-05
feat(web): redesign /logs with LIVE/PAUSED, search, autoscroll

### LINKED ISSUE #63: feat(web): redesign Logs page with LIVE/PAUSED indicator, search, autoscroll
## Context

`LogsPanel.tsx` streams CloudWatch lines via SSE into a panel at the bottom of the dashboard. Pause/Resume is a button label only — there's no big LIVE/PAUSED indicator, no in-stream search, no log-level filter, no autoscroll toggle.

## Scope

**New `/logs` route** with:

- **LIVE / PAUSED status badge** (top-right of panel): pulsing cyan when live, muted slate when paused. Replace the bare button with a clear pill + a separate Pause/Resume button.
- **Game selector** (shadcn `<Select>` with search) — replaces the plain `<select>`.
- **Color-coded log levels**: detect `INFO` / `WARN` / `ERROR` / `DEBUG` patterns in each line and render a small badge per line (cyan/orange/red/slate). Fallback to plain when no level can be detected.
- **In-stream search**: text input above the stream that highlights matches in the visible buffer (does not affect the stream).
- **Level filter**: multi-select to hide levels (default all on).
- **Autoscroll toggle**: on by default; turning off freezes scroll position when new lines arrive.
- **Line count + age** in the footer ("142 lines · oldest 4m ago").

Buffering, SSE handling, and the Pause/Resume *behavior* should not change — only the surrounding UX.

## Acceptance criteria

- [ ] `/logs` route reachable; `<LogsPanel>` removed from dashboard.
- [ ] LIVE/PAUSED badge updates correctly when pausing/resuming.
- [ ] Game selector switches the stream and resets the buffer.
- [ ] Search highlights matches without filtering them out (filter is a separate control).
- [ ] Autoscroll behavior matches spec (on = follow tail, off = freeze).
- [ ] Existing SSE plumbing in `useFileManager`/`LogsPanel` left functionally intact.

## Dependencies

- Part of #57.
- **Blocked by** #58, #59.

CoderCoco:
Closes #63

Promote the bottom-of-dashboard log tail into a dedicated `/logs` route with
the ops-dashboard treatment:

- LIVE/PAUSED status pill (pulsing cyan / muted slate) replaces the bare
  Pause button label; the button itself becomes a separate Pause/Resume.
- Searchable game selector (custom GameCombobox) replaces the plain `<select>`
  and resets the buffer on switch.
- Per-line color-coded level badges (INFO/WARN/ERROR/DEBUG) detected via a
  word-boundary regex; lines without a level fall back to plain text.
- In-stream search input highlights matches in the visible buffer using
  `<mark>` spans without filtering them out.
- Multi-select level filter (DropdownMenu + Badge) hides whole levels;
  defaults to all on.
- Autoscroll toggle, on by default; turning it off freezes the scroll
  position even as new lines arrive.
- Footer summary: line count, oldest-line age, hidden-level count, and
  buffered-while-paused count.

SSE plumbing (initial /api/logs snapshot + EventSource on
/api/logs/:game/stream) and the pause-buffers/resume-flushes behaviour are
preserved verbatim from the old `LogsPanel`. Removed `LogsPanel` from the
Dashboard and deleted the file — `/logs` is now the only consumer.

https://claude.ai/code/session_01AJDEL37XXB5i6GRDJugNvx


app/packages/web/src/pages/DashboardPage.tsx
```diff
@@ -6,15 +6,13 @@ import { api, type ActualCosts } from '../api.js';
...
 
 /**
  * Dashboard route (`/`) — top KPI strip, then a search-filterable grid of
- * GameCards, then the logs panel (which will move to its own route in
- * CoderCoco/game-server-deploy#63). Cost analysis lives at `/costs`,
- * Discord settings at `/discord`, and the watchdog at `/settings`. The
- * search input narrows the grid by game name or hostname client-side.
+ * GameCards. Cost analysis lives at `/costs`, Discord settings at `/discord`,
+ * the live log tail at `/logs`, and the watchdog at `/settings`. The search
+ * input narrows the grid by game name or hostname client-side.
```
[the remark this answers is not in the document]
CoderCoco: Renamed the PR title to `feat(web): redesign /logs with LIVE/PAUSED, search, autoscroll` (62 chars) so the squash subject lands under the ~70-char guidance from `CLAUDE.md`.

---

app/packages/web/src/components/GameCombobox.tsx
```diff
@@ -0,0 +1,110 @@
...
+      >
+        <span className="truncate font-[var(--font-mono)]">{value || 'Select a game…'}</span>
+        <ChevronsUpDown className="h-4 w-4 opacity-60" />
+      </button>
+
+      {open && (
+        <div
+          role="listbox"
+          className="absolute z-50 mt-1 w-72 rounded-[var(--radius-md)] border border-[var(--color-border)] bg-[var(--color-surface)] shadow-lg"
+        >
```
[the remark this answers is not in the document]
CoderCoco: Good catch — fixed in 59e8015. The trigger no longer claims `aria-haspopup="listbox"` and the popover container no longer has `role="listbox"`, so we don't lie to screen readers about an interaction model we don't actually implement (no `role="option"` items, no arrow-key nav). Replaced with a dynamic `aria-label` (`"Game selector, ${value} selected"`) so the trigger's purpose + current value are still announced. The items stay as plain `<button>` elements — accurate for what is really a popover with a search input + clickable buttons.

I considered the alternative path (full listbox semantics with `role="option"`, `aria-selected`, and arrow-key focus management) but that's a meaningfully larger change and outside the scope of #63; happy to do it as a follow-up if a richer keyboard story is wanted.

---

app/packages/web/e2e/specs/logs.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+  test('should render LIVE badge and seeded log lines', async ({ authedPage: page }) => {
+    await stubApis(page, {
+      statuses: [{ game: 'minecraft', state: 'stopped' }],
+      logLines: { minecraft: SAMPLE_LOG_LINES },
+    });
+    await page.goto('/logs');
+
+    await expect(page.getByRole('heading', { name: 'Server Logs' })).toBeVisible();
+    await expect(page.getByText('Live', { exact: true })).toBeVisible();
+    await expect(page.getByText('Server started on port 25565')).toBeVisible();
```
CoderCoco: These should be part of the page object

app/packages/web/src/components/GameCombobox.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const onKeyDown = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') setOpen(false);
+    };
+    document.addEventListener('mousedown', onMouseDown);
+    document.addEventListener('keydown', onKeyDown);
+    // Focus the search input on open
+    inputRef.current?.focus();
+    return () => {
+      document.removeEventListener('mousedown', onMouseDown);
+      document.removeEventListener('keydown', onKeyDown);
```
CoderCoco: Is there a reason why we are adding event listeners to the document instead of react containers?

app/packages/web/src/pages/LogsPage.tsx
CoderCoco, line 1: Missing unit tests for this file

app/packages/web/src/components/GameCombobox.tsx
CoderCoco, line 1: Missing unit tests for this file

CoderCoco (commented):
Ensure that we sare using page objects for retrieving elements on the page and it looks like there is a lack of unit testing for individual react files. Can we update the claude.md to add this context?

app/packages/web/e2e/specs/logs.spec.ts
CoderCoco: Done in 99f2e64. New `e2e/pages/LogsPage.ts` follows the established convention from `DashboardPage` / `CostsPage` / `AppLayout`, wrapping every locator the spec used to reach for: LIVE/PAUSED pill, game combobox + popover search, in-stream search, Levels multi-select, Autoscroll toggle, Pause/Resume button, per-line level badges, highlight `<mark>`s, and the footer summary. Added a `logs` Playwright fixture so specs read `async ({ logs }) => ...`. All 7 logs specs migrated and still passing — same assertions, no raw `page.getByX(...)` for page-owned UI.

---

app/packages/web/src/components/GameCombobox.tsx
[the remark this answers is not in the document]
CoderCoco: Follow-up: also extracted the document-listener wiring into a reusable `useClickOutside(ref, onClose, enabled)` hook in 9828f29 so the rationale lives in one place (TSDoc on the hook) and any future popover-style component can reuse it. The hook still attaches to `document` — that part fundamentally has to be document-level, since synthetic-event delegation only sees clicks rooted in the component tree — but consumers now hand in just a ref + callback.

---

app/packages/web/src/components/GameCombobox.tsx
CoderCoco: Refactored in 9828f29. Document-level listeners are the standard React pattern for "click outside this subtree" because synthetic-event delegation only sees clicks rooted in the component tree — there's no React-native primitive for "any click anywhere on the page that wasn't inside this component". Backdrop overlays solve the same problem at the cost of a stacking-context layer; portal-based UI libraries like Radix do this in their own internals.

Pulled the wiring out of `GameCombobox` into a small reusable `useClickOutside(ref, onClose, enabled)` hook in `app/packages/web/src/hooks/useClickOutside.ts` so the rationale lives in one place (TSDoc on the hook) and the next popover-style component we add can reuse it. The hook is a no-op when `enabled` is false so consumers can keep the call unconditional and let React strip the listeners while the popover is closed.

---

app/packages/web/src/pages/LogsPage.tsx
CoderCoco: Added in 1966b4c. New `app/packages/web/src/pages/LogsPage.test.tsx` (8 specs) covers: the LIVE/PAUSED badge, seeded log lines, per-line level badges (INFO/WARN/ERROR/DEBUG), the search-input `<mark>` highlighting without filtering lines out, the Levels multi-select hiding ERROR lines, the Pause/Resume toggle round-trip, the footer summary, and that `api.logs(game)` is called with the selected game on mount.

Same commit pulls in the missing infra — there was no React unit-test setup in the repo at all (Vitest ran in the `node` environment, glob covered only `*.test.ts`). It now adds `@testing-library/react`, `@testing-library/user-event`, `@testing-library/jest-dom`, and `jsdom` as devDeps; flips `packages/web/**` to the `jsdom` environment via `environmentMatchGlobs` (server tests stay on Node); and registers jest-dom matchers + `afterEach(cleanup)` in `app/vitest.setup.ts` (we run `globals: false` so RTL's auto-cleanup needs to be wired explicitly). Documented the convention in `CLAUDE.md` so future PRs know where to put React component tests and what to cover.

---

app/packages/web/src/components/GameCombobox.tsx
CoderCoco: Added in 1966b4c. New `app/packages/web/src/components/GameCombobox.test.tsx` (8 specs) covers: trigger label rendering with the selected value, the placeholder when no game is selected, the `aria-expanded` collapsed-by-default state, opening the popover and revealing every game, search-input filtering, the empty-state when nothing matches, the `onChange` callback firing exactly once with the right argument, and Escape-to-close.

Same commit pulls in the missing infra (Vitest + jsdom + RTL + user-event) and documents the convention in `CLAUDE.md` — see the reply on `LogsPage.tsx` for the longer rationale.

---

app/packages/web/e2e/pages/LogsPage.ts
```diff
@@ -0,0 +1,166 @@
...
+   * so a `/Levels/` regex matches no matter how many levels are currently
+   * shown — narrow with `levelsTriggerWithCount` for an exact count.
+   */
+  levelsTrigger(): Locator {
+    return this.page.getByRole('button', { name: /Levels/ });
+  }
+
+  /** Levels trigger asserted to display a specific visible-count (e.g. `3/4`). */
+  levelsTriggerWithCount(visible: number, total = 4): Locator {
+    const escaped = `${visible}/${total}`.replace(/\//g, '\\/');
```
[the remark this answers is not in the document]
CoderCoco: Fixed in f80dce6. The flagged code constructed a dynamic regex (`Levels.*${count}/${total}`) and only escaped forward slashes — backslashes and other regex metacharacters weren't handled. The inputs are typed `number`, so an exploit isn't reachable in practice, but switched to `getByRole('button', { name: 'Levels (V/T)', exact: true })` which doesn't construct a regex at all and clears the alert outright.

---

PR #296, opened 2026-07-18
feat(desktop-main): TerraformService.plan() producing tfplan artifact

CoderCoco:
Closes #173

## Summary

- Adds `ConfigService.getRunsDir()` to resolve `userData/runs` (creating it on demand), matching the existing pattern for `getTerraformDir()`.
- Extracts the streaming spawn pump previously inlined in `init()` into a shared private `spawnAndStream()` generator helper on `TerraformService`, so `plan()` can reuse the same abort-signal handling and chunked stdout/stderr streaming.
- Implements `TerraformService.plan(runId, tfvarsVersionId)` — pulls tfvars from the `RemoteFileStore`, writes it to `userData/runs/<runId>/terraform.tfvars.json`, and spawns `terraform plan -var-file=... -out=<runId>.tfplan`, scanning stdout for the plan summary line (add/change/destroy counts) while streaming raw output via the shared generator.
- Fixes an abort-listener leak: when `plan()`'s generator is terminated early (consumer `break`/`.return()`/`.throw()`), it now explicitly finalizes the inner `spawnAndStream()` generator via `.return()` so its `finally` block (which removes the abort listener) always runs.

## Changes

```
 .../desktop-main/src/modules/terraform.module.ts   |  16 +-
 .../src/services/ConfigService.test.ts             |  21 +
 .../desktop-main/src/services/ConfigService.ts     |  36 ++
 .../src/services/TerraformService.init.test.ts     | 436 ++++++++++++++-
 .../src/services/TerraformService.plan.test.ts     | 608 +++++++++++++++++++++
 .../src/services/TerraformService.test.ts          |  48 +-
 .../desktop-main/src/services/TerraformService.ts  | 495 ++++++++++++++---
 7 files changed, 1529 insertions(+), 131 deletions(-)
```

## Test plan

- [x] `plan()` writes the pulled tfvars to `userData/runs/<runId>/terraform.tfvars.json` before spawning — covered by `TerraformService.plan.test.ts`.
- [x] `plan()` spawns `terraform plan -var-file=... -out=<runId>.tfplan`, persisting the tfplan artifact under `userData/runs/<runId>/` — covered by `TerraformService.plan.test.ts`.
- [x] Plan summary (add/change/destroy counts) is parsed from stdout and returned alongside the artifact path — covered by `TerraformService.plan.test.ts`.
[...]

PR #297, opened 2026-07-18
feat(desktop-main): TerraformService.apply() with stale-plan guard

### LINKED ISSUE #175: feat(desktop-main): TerraformService.apply() with stale-plan guard
Child of #138 (epic: local terraform orchestration). Authoritative scope in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Scope

`apply(runId, tfvarsVersionId, planFile)` refuses if the cached tfvars version differs from `tfvarsVersionId` (surface `StalePlanError`). Otherwise spawns `terraform apply <planFile>`. Streams output. Persists a run record on completion.

## Acceptance

- Stale-plan condition surfaces an inline error in the renderer.
- Successful apply writes a `runs` table row and clears the pending-changes banner.

CoderCoco:
Closes #175

## Summary

- Adds `TerraformService.apply(runId, tfvarsVersionId, planFile, signal)` — spawns `terraform apply -input=false -no-color <planFile>` inside the terraform working directory, streaming output chunk-by-chunk (mirroring `init()`/`plan()`'s shape) and resolving with `{ runId, added, changed, destroyed }` parsed from Terraform's `Apply complete!` summary line.
- Adds a stale-plan guard (`assertPlanTfvarsNotStale`): when `tfvarsVersionId` is supplied and the tfvars source is S3-backed, compares it against the head version of the tfvars object via `remoteFileStore.listVersions()` *before* spawning `terraform`, throwing the new `StalePlanError` if the remote tfvars changed since the plan was generated. No-op in local-file mode or when `tfvarsVersionId` is omitted.
- Persists a `TerraformRunRecord` to `<runsDir>/<runId>/run.json` once the spawned process closes (clean exit, non-zero exit, or aborted via `signal`) — `runId`, `kind: 'apply'`, `startedAt`/`completedAt`, `exitCode` (`null` when aborted), and `tfvarsVersionId` — a lightweight local run history for a future Apply-history UI.
- Extends the shared `workspaceInFlight` lock (previously `'init' | 'plan'`) to include `'apply'`, since apply mutates the same `backend/.terraform` workspace state that `init`/`plan` read — overlapping runs would race.
- New error types: `StalePlanError` (pre-spawn staleness guard) and `TerraformApplyError` (non-zero exit code from the spawned process), both distinct from the existing `TerraformNotFoundError`.
- Rejects synchronously (before spawning) if `planFile` doesn't exist on disk, or if another `apply()`/`plan()`/`init()` is already in flight.

## Changes

```
 .../src/services/TerraformService.apply.test.ts    | 772 +++++++++++++++++++++
 .../desktop-main/src/services/TerraformService.ts  | 324 ++++++++-
 2 files changed, 1085 insertions(+), 11 deletions(-)
```

## Test plan

- [x] Stale-plan condition (`tfvarsVersionId` no longer matches the S3 tfvars head version) throws `StalePlanError` before `terraform` is spawned — covered by `TerraformService.apply.test.ts`.
- [x] Successful apply spawns `terraform apply -input=false -no-color <planFile>`, streams output, and returns `{ runId, added, changed, destroyed }` parsed from the `Apply complete!` summary line.
- [x] A run record (`run.json`) is written to `<runsDir>/<runId>/` on completion — for clean exit, non-zero exit, and abort-via-signal paths.
- [x] Non-zero exit code throws `TerraformApplyError`; the run record still records the `exitCode`.
- [x] Abort signal kills the spawned process, ends the generator cleanly (no throw), and records `exitCode: null`.
- [x] `apply()` shares the `workspaceInFlight` lock with `init()`/`plan()` — rejects synchronously if any of the three is already running.
- [x] Missing `planFile` on disk rejects synchronously without spawning.
- [x] Early generator termination (consumer `break`/`.return()`/`.throw()`) finalizes the inner `spawnAndStream()` generator so the abort listener is always removed (no leak).
- [x] `npm run app:test` passing for the touched workspace (desktop-main).
- [x] `npm run app:lint` clean.

### DIFF
--- a/app/packages/desktop-main/src/services/TerraformService.apply.test.ts
+++ b/app/packages/desktop-main/src/services/TerraformService.apply.test.ts
@@ -1,0 +1,772 @@
+import { EventEmitter } from 'node:events';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+/*
+ * Spy variables must be hoisted before vi.mock() factories run, because
+ * vi.mock() calls are lifted to the top of the compiled output above regular
+ * declarations.
+ */
+const { execFileMock, spawnMock, mkdirSyncMock, writeFileSyncMock, existsSyncMock } = vi.hoisted(() => {
+  const execFileMock = vi.fn();
+  const spawnMock = vi.fn();
+  const mkdirSyncMock = vi.fn();
+  const writeFileSyncMock = vi.fn();
+  const existsSyncMock = vi.fn();
+  return { execFileMock, spawnMock, mkdirSyncMock, writeFileSyncMock, existsSyncMock };
+});
+
+vi.mock('node:child_process', () => ({
+  execFile: execFileMock,
+  spawn: spawnMock,
+}));
+
+vi.mock('node:fs', () => ({
+  mkdirSync: mkdirSyncMock,
+  existsSync: existsSyncMock,
+  writeFileSync: writeFileSyncMock,
+  copyFileSync: vi.fn(),
+}));
+
+import {
+  TerraformService,
+  TerraformNotFoundError,
+  TerraformApplyError,
+  StalePlanError,
+  type TerraformInitConfig,
+  type TerraformRunChunk,
+  type TerraformApplyResult,
+  type TerraformRunRecord,
+} from './TerraformService.js';
+import type { ConfigService } from './ConfigService.js';
+import type { RemoteFileStore } from '@hyveon/shared';
+
+/** Error-first callback shape `util.promisify` invokes the mocked `execFile` with. */
+type ExecFileCallback = (error: Error | null, result?: { stdout: string; stderr: string }) => void;
+
+/**
+ * Extracts the error-first callback from an `execFile` call's arguments,
+ * regardless of whether `util.promisify` invoked it with or without an
+ * `options` object.
+ */
+function lastArgAsCallback(args: unknown[]): ExecFileCallback {
+  return args[args.length - 1] as ExecFileCallback;
+}
+
+/** Queues a successful `execFile` invocation (stdout/stderr) for the next call. */
+function queueExecFileSuccess(stdout: string, stderr = ''): void {
+  execFileMock.mockImplementationOnce((...args: unknown[]) => {
+    lastArgAsCallback(args)(null, { stdout, stderr });
+  });
+}
+
+/** Queues a failing `execFile` invocation (e.g. binary not found) for the next call. */
+function queueExecFileFailure(error: Error = new Error('spawn ENOENT')): void {
+  execFileMock.mockImplementationOnce((...args: unknown[]) => {
+    lastArgAsCallback(args)(error);
+  });
+}
+
+/** Queues a successful binary lookup followed by a successful `-json` version response. */
+function queueSuccessfulResolution(binaryPath = '/usr/local/bin/terraform', version = '1.7.0'): void {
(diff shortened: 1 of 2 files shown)


CoderCoco:
I've addressed the following feedback:

- `app/packages/desktop-main/src/services/TerraformService.apply.test.ts:327` — driveChild test helper fires stdout+stderr+close all synchronously before the first yielded chunk is awaited, so the streaming test would pass even with a buffer-until-exit implementation.
- `app/packages/desktop-main/src/services/TerraformService.ts:642` — runId is unchecked before being joined into runsDir paths and planFile isn't verified to correspond to that runId, allowing path traversal or applying an unrelated plan artifact.
- `app/packages/desktop-main/src/services/TerraformService.ts:721` — When apply()'s generator is force-closed (consumer break/.return()/.throw()), the child is killed via spawnAndStream's finally but writeRunRecord (which lives after the inner try/finally block) never executes, so the cancelled run is never persisted.
- `app/packages/desktop-main/src/services/TerraformService.ts:727` — spawnAndStream determines aborted:true by reading signal.aborted after the loop exits, so an abort that fires after the process has already closed cleanly can misclassify a successful apply as aborted (exitCode recorded as null).
- `app/packages/desktop-main/src/services/TerraformService.ts:732` — writeRunRecord's mkdirSync/writeFileSync calls aren't wrapped, so a filesystem error while persisting run.json throws and replaces a successful result, a clean abort, or a TerraformApplyError with an unrelated persistence exception.

Please re-review when you get a chance.

CoderCoco:
I've addressed the following feedback:

- `app/packages/desktop-main/src/services/TerraformService.ts:875` — CodeRabbit: apply()'s outer finally infers cancellation from a missing run-record write, but that same path also runs when spawnAndStream throws or an unrelated error occurs after the child already finished, mislabeling failure/success runs as cancelled.
- `app/packages/desktop-main/src/services/TerraformService.ts:1099` — CodeRabbit review summary: the abort latch in spawnAndStream still accepts an abort signal that fires after the child has already closed, retroactively misclassifying a completed run as aborted; onAbort should guard with `if (closed) return;`.

Please re-review when you get a chance.

CoderCoco:
I've addressed the following feedback:

- `app/packages/desktop-main/src/services/TerraformService.ts:1195` — In spawnAndStream's finally block, `closed === false` is treated as proof the child is still live before calling onForceKill/child.kill(), but the process may have already exited while stdio streams are still draining (before the 'close' event fires), which would mislabel a completed run as force-killed/cancelled.

Please re-review when you get a chance.

CoderCoco:
Thanks for the rollup. The closed===false/onForceKill finding is the same one tracked on the inline thread at TerraformService.ts:1195 — leaving that one open there rather than duplicating here. On the two outside-diff-range points: the abort listener is already attached synchronously immediately after spawn, in the same tick as the pre-spawn signal.aborted checks in apply() (no await intervenes before addEventListener), so there's no registration gap to close. The optional tfvarsVersionId behavior in assertPlanTfvarsNotStale (skip the staleness check when it's omitted) mirrors the identical, already-merged pattern in plan()'s pullVarFile from PR #296 — it's intentional, existing design (enforcing 'must supply a version when S3-backed' would be a behavior change belonging to the calling layer, not this PR's scope), not a new regression introduced here.

PR #315, opened 2026-07-24
feat(terraform): replace ALB with in-task Caddy TLS sidecar

CoderCoco:
Part of #292

## Summary

Replaces the always-on ALB fronting HTTPS games (~$16-17/mo purely for existing, even at zero tasks running — the one resource in the stack that breaks the "pay nothing when idle" design) with a Caddy reverse-proxy **sidecar container** added directly to the HTTPS game's own Fargate task.

- `aws_ecs_task_definition.game`: conditionally adds a version-pinned `caddy:2-alpine` sidecar (443/80, `essential = true`) plus a cert-storage EFS volume for every `https = true` game, driven entirely by the existing `game_servers` `for_each` — no hand-written per-game resources.
- Caddy obtains and renews certificates via Let's Encrypt automatic HTTPS (HTTP-01/TLS-ALPN-01); issued certs persist on a new dedicated `{game}-certs` EFS access point so restarts reuse the cert instead of re-issuing.
- Security group: HTTPS games get public 443/80 ingress for the sidecar; their raw container port loses public ingress entirely (the sidecar reaches it over `localhost` inside the task).
- **Deletes `terraform/aws/alb.tf` outright** — ALB, both listeners, target groups, listener rules, ALB security group, ACM certificate + validation, and the static Route 53 alias records.
- HTTPS games return to the same Lambda-managed DNS lifecycle (`update-dns` UPSERT/DELETE) every other game already uses. Trims `HTTPS_GAMES`/`ALB_TARGET_GROUPS` env vars and the `elasticloadbalancing:*` IAM statement from both the `update-dns` and `watchdog` Lambda Terraform wiring — the deployed Lambda code already treats these as optional (defaults to empty via `?? ''`/`?? '{}'`), so nothing breaks before the dead-code-removal follow-up PR lands.
- Removes the `acm_certificate_domain` variable (full 5-point Terraform-variable checklist: both `variables.tf` copies, `main.tf` passthrough, `terraform.tfvars.example`, `docs/docs/components/terraform.md`) and the `alb_dns_name`/`acm_certificate_arn` outputs (both `outputs.tf` copies).
- Updates `docs/docs/components/terraform.md` and `docs/docs/setup.md` for the new `https` semantics and first-boot ACME issuance timing.

## Migration note

**The `terraform apply` that lands this must run with any HTTPS game(s) stopped** — it deletes the static alias record and the ALB while producing a new task-definition revision, so a task already running would keep using the old revision with no ALB in front of it.

**The IAM deploy-policy trim (`elasticloadbalancing:*`, `acm:*` in `docs/docs/setup.md`) is deferred to a follow-up PR** — the destroy apply that removes the ALB stack still needs those permissions. Trimming them here would break the very apply this PR enables.

## Test plan

- [x] `terraform fmt -check -recursive` — clean
- [x] `terraform validate` — success
- [x] `tflint` (AWS ruleset) — no findings
- [ ] `terraform apply` against live AWS with the HTTPS game stopped (tracked separately — see migration note above; not run as part of this PR)
- [ ] Post-apply: verify `https://{game}.{zone}` serves a valid cert within ~2 min of RUNNING, and a restart reuses the cert (no new ACME order)


CoderCoco:
@coderabbitai review

CoderCoco:
## Fixes Applied Successfully

Fixed 3 file(s) based on 2 CodeRabbit feedback item(s).

**Files modified:**
- `docs/docs/setup.md`
- `terraform/aws/variables.tf`
- `terraform/variables.tf`

**Commit:** `457ba70`

The latest autofix changes are on the `claude/issue-292-caddy-sidecar-terraform` branch.

CoderCoco:
## Fixes Applied Successfully

Fixed 2 file(s) based on 3 CodeRabbit feedback item(s) (one duplicate finding bundled with its pair; one earlier finding confirmed already resolved by commit 457ba70's prior validation fix — no code change needed for that one).

**Files modified:**
- `terraform/aws/variables.tf`
- `terraform/variables.tf`

Extended the `game_servers` HTTPS validation to require exact-case `protocol = "tcp"` (matching what ECS actually requires, instead of a case-insensitive check that could pass validation and then fail `terraform apply`), and to reject any declared port using 80 or 443 on an HTTPS game (those collide with the Caddy sidecar's own host port bindings in awsvpc mode).

**Commit:** `04fd592`

The latest autofix changes are on the `claude/issue-292-caddy-sidecar-terraform` branch.

CoderCoco:
@coderabbitai all fixess addressed. can you review or approve?

CoderCoco:
## Fixes Applied Successfully

Fixed 1 finding out of 3 reported this round (the other 2 were verified as already resolved / not actually reproducible and left as-is — see thread discussion).

**Files modified:**
- `terraform/aws/variables.tf`
- `terraform/variables.tf`

Extended the HTTPS `game_servers` validation so every declared port (not just `ports[0]`) is checked against valid ECS protocol values (`tcp`/`udp`), preventing a garbage protocol string on a non-first port from slipping past validation and failing `terraform apply`.

**Commit:** `be75bc6`

The latest autofix changes are on the `claude/issue-292-caddy-sidecar-terraform` branch.

Issue #211, opened 2026-05-10
feat(web): Reconfigure entry point in Settings


CoderCoco:
Child of #139 (epic: first-run wizard + credentials UX). Authoritative scope in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Scope

Settings page surfaces a "Reconfigure" button that re-launches the wizard against the existing electron-store state. Skips steps the operator has already completed, with an "Edit" affordance per step.

## Acceptance

- Reconfigure preserves existing config except for the field the operator changes.
- Survives mid-flow cancel.

PR #329, opened 2026-07-26
feat(web): terraform-init wizard step and completion flow

### LINKED ISSUE #210: feat(web): wizard step — terraform init with live log
Child of #139 (epic: first-run wizard + credentials UX). Authoritative scope in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Scope

Final wizard step: invokes `TerraformService.init()` (D.2) and renders the live IPC log stream. Completion button enabled on exit-0; surfaces error UI on non-zero exit.

## Acceptance

- First-Run completes end-to-end on a clean machine.
- Output ANSI colors render correctly.
- Completion button only enables on exit code 0.

CoderCoco:
Closes #210

## Summary
- `terraform-init-step.component.tsx` — the wizard's fifth and final step. Consumes the existing `gsd.terraform.init(backendConfig, signal)` async iterable, rendering live ANSI-colored output via the already-shipped `AnsiLogViewer`. "Finish setup" enables only once the run completes without throwing; a failed run (the iterable throwing) shows the captured log plus a Retry button that re-invokes `gsd.terraform.init`.
- `FirstRunWizardService` (new) — owns the wizard's resumable step-progress file (`userData/wizard-state.json`, per design.md decision 4), mirroring `ConfigService.readUserDataPath`'s Electron-vs-tmpdir seam. `getProgress()`/`recordStep()` degrade to the `prerequisites` step on any missing/corrupt/unrecognized state rather than throwing; `complete()` sets `wizardCompleted: true` in `ElectronStoreService`.
- Three new `wizard.*` IPC patterns + preload/`gsd-api.ts` mirrors: `wizard.progress.get`, `wizard.progress.save`, `wizard.complete`.
- Wizard shell: adds `terraform-init` as the fifth `WIZARD_STEPS` entry, computes `backendConfig` (bucket/region/dynamodbTable) from the bootstrap step's resource names + the credentials step's region, resumes to the last-recorded step on mount, and persists the current step via `wizard.progress.save` on every step change (fire-and-forget — a failed write only degrades resume, never blocks the wizard). The shared Next button is hidden on this last step in favor of the step's own Finish button.
- `app.component.tsx`: `FirstRunWizard` now takes an `onComplete` prop; `App` flips straight to the routed dashboard once it fires, without waiting on another `wizard.state.get` round-trip.

## Test plan
- [x] `npm run app:test` — 2080/2080 passing (new: `FirstRunWizardService` specs covering resume + corrupt-file fallback + `complete`; `WizardController` pattern-registration/delegation specs for the three new channels; `TerraformInitStep` RTL specs — streaming render, ANSI colors, enable-on-success, error+retry, Finish calling `wizard.complete`; wizard-shell specs for hiding Next on the last step, per-step `wizard.progress.save`, end-to-end Finish → `onComplete`, and resume-on-mount incl. a rejected `getProgress` degrading safely)
- [x] `npm run app:lint` — 0 errors (3 pre-existing warnings in an untouched file)
- [x] `npm run app:build` (full `tsc -b` + `vite build`) — clean

### DIFF
--- a/app/packages/desktop-main/src/services/FirstRunWizardService.test.ts
+++ b/app/packages/desktop-main/src/services/FirstRunWizardService.test.ts
@@ -1,0 +1,96 @@
+import 'reflect-metadata';
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import { mkdtempSync, mkdirSync, rmSync, writeFileSync, readFileSync, existsSync } from 'fs';
+import { join } from 'path';
+import { tmpdir } from 'os';
+import { FirstRunWizardService } from './FirstRunWizardService.js';
+import type { ElectronStoreService } from './ElectronStoreService.js';
+
+/** Test-only subclass exposing the protected `stateFilePath` seam so tests can point it at a real scratch directory. */
+class TestableFirstRunWizardService extends FirstRunWizardService {
+  public override stateFilePath(): string {
+    return super.stateFilePath();
+  }
+}
+
+/** Build an `ElectronStoreService` stub whose `set()` calls are observable. */
+function makeStore(): ElectronStoreService {
+  return { set: vi.fn() } as Partial<ElectronStoreService> as ElectronStoreService;
+}
+
+describe('FirstRunWizardService', () => {
+  let scratchDir: string;
+  let statePath: string;
+  let service: TestableFirstRunWizardService;
+  let store: ElectronStoreService;
+
+  beforeEach(() => {
+    scratchDir = mkdtempSync(join(tmpdir(), 'wizard-state-test-'));
+    statePath = join(scratchDir, 'nested', 'wizard-state.json');
+    mkdirSync(join(scratchDir, 'nested'));
+    store = makeStore();
+    service = new TestableFirstRunWizardService(store);
+    vi.spyOn(service, 'stateFilePath').mockReturnValue(statePath);
+  });
+
+  afterEach(() => {
+    rmSync(scratchDir, { recursive: true, force: true });
+  });
+
+  describe('getProgress', () => {
+    it('should default to the prerequisites step when the state file does not exist', async () => {
+      const progress = await service.getProgress();
+      expect(progress).toEqual({ step: 'prerequisites' });
+    });
+
+    it('should return the persisted step when the state file holds a recognized step name', async () => {
+      writeFileSync(statePath, JSON.stringify({ step: 'bootstrap' }), 'utf-8');
+      const progress = await service.getProgress();
+      expect(progress).toEqual({ step: 'bootstrap' });
+    });
+
+    it('should default to the prerequisites step when the file contains invalid JSON', async () => {
+      writeFileSync(statePath, 'not valid json{{{', 'utf-8');
+      const progress = await service.getProgress();
+      expect(progress).toEqual({ step: 'prerequisites' });
+    });
+
+    it('should default to the prerequisites step when the file holds an unrecognized step name', async () => {
+      writeFileSync(statePath, JSON.stringify({ step: 'some-future-step' }), 'utf-8');
+      const progress = await service.getProgress();
+      expect(progress).toEqual({ step: 'prerequisites' });
+    });
+  });
+
+  describe('recordStep', () => {
+    it('should write the given step to the state file, creating parent directories as needed', async () => {
+      expect(existsSync(statePath)).toBe(false);
+
+      await service.recordStep('credentials');
+
+      expect(JSON.parse(readFileSync(statePath, 'utf-8'))).toEqual({ step: 'credentials' });
+    });
+
+    it('should be resumable: a later getProgress reflects the step recorded by an earlier recordStep', async () => {
+      await service.recordStep('terraform-init');
+
+      const progress = await service.getProgress();
+
+      expect(progress).toEqual({ step: 'terraform-init' });
+    });
+
+    it('should overwrite a previously recorded step', async () => {
+      await service.recordStep('pick-cloud');
+      await service.recordStep('bootstrap');
+
+      expect(await service.getProgress()).toEqual({ step: 'bootstrap' });
+    });
+  });
+
+  describe('complete', () => {
+    it('should set wizardCompleted to true in ElectronStoreService', () => {
+      service.complete();
+      expect(store.set).toHaveBeenCalledWith('wizardCompleted', true);
+    });
+  });
+});
--- a/app/packages/desktop-main/src/services/FirstRunWizardService.test.ts
+++ b/app/packages/desktop-main/src/services/FirstRunWizardService.test.ts
@@ -18,7 +18,8 @@ class TestableFirstRunWizardService extends FirstRunWizardService {
 
 /** Build an `ElectronStoreService` stub whose `set()` calls are observable. */
 function makeStore(): ElectronStoreService {
-  return { set: vi.fn() } as Partial<ElectronStoreService> as ElectronStoreService;
+  const store: Partial<ElectronStoreService> = { set: vi.fn() };
+  return store as ElectronStoreService;
 }
 
 describe('FirstRunWizardService', () => {
@@ -72,6 +73,14 @@ describe('FirstRunWizardService', () => {
       await service.recordStep('credentials');
 
       expect(JSON.parse(readFileSync(statePath, 'utf-8'))).toEqual({ step: 'credentials' });
+    });
+
+    it('should reject a step name outside WIZARD_STEPS rather than writing it', async () => {
+      // The IPC boundary erases WizardStepName's compile-time guarantee, so a
+      // malformed/unexpected value must be rejected explicitly at runtime.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any -- deliberately bypassing the compile-time type to exercise the runtime guard
+      await expect(service.recordStep('not-a-real-step' as any)).rejects.toThrow('Unsupported wizard step');
+      expect(existsSync(statePath)).toBe(false);
     });
 
     it('should be resumable: a later getProgress reflects the step recorded by an earlier recordStep', async () => {
--- a/app/packages/desktop-main/src/services/FirstRunWizardService.test.ts
+++ b/app/packages/desktop-main/src/services/FirstRunWizardService.test.ts
@@ -6,10 +6,13 @@ import { tmpdir } from 'os';
 import { FirstRunWizardService } from './FirstRunWizardService.js';
 import type { ElectronStoreService } from './ElectronStoreService.js';
 
-/** Test-only subclass exposing the protected `stateFilePath` seam so tests can point it at a real scratch directory. */
+/** Test-only subclass exposing the protected `stateFilePath`/`userDataPath` seams so tests can point them at a real scratch directory or assert their composition directly. */
 class TestableFirstRunWizardService extends FirstRunWizardService {
   public override stateFilePath(): string {
     return super.stateFilePath();
+  }
+  public override userDataPath(): string {
+    return super.userDataPath();
   }
 }
 
@@ -88,9 +91,38 @@ describe('FirstRunWizardService', () => {
   });
 
   describe('complete', () => {
-    it('should set wizardCompleted to true in ElectronStoreService', () => {
-      service.complete();
+    it('should set wizardCompleted to true in ElectronStoreService', async () => {
+      await service.complete();
       expect(store.set).toHaveBeenCalledWith('wizardCompleted', true);
+    });
+
+    it('should clear the resume file, so a future re-entry (e.g. Settings Reconfigure) starts clean', async () => {
+      await service.recordStep('terraform-init');
+      expect(existsSync(statePath)).toBe(true);
+
+      await service.complete();
+
+      expect(existsSync(statePath)).toBe(false);
+      expect(await service.getProgress()).toEqual({ step: 'prerequisites' });
+    });
+
+    it('should not throw when the resume file never existed', async () => {
+      await expect(service.complete()).resolves.toBeUndefined();
+    });
+  });
+
+  describe('userDataPath', () => {
+    it('should compose stateFilePath from userDataPath and the fixed filename', () => {
+      const unspiedService = new TestableFirstRunWizardService(makeStore());
+      vi.spyOn(unspiedService, 'userDataPath').mockReturnValue('/fake/userData');
+
+      expect(unspiedService.stateFilePath()).toBe(join('/fake/userData', 'wizard-state.json'));
+    });
+
+    it('should fall back to a hyveon-wizard-namespaced OS temp directory outside Electron', () => {
+      const unspiedService = new TestableFirstRunWizardService(makeStore());
+
+      expect(unspiedService.userDataPath()).toBe(join(tmpdir(), 'hyveon-wizard'));
     });
   });
 });
--- a/app/packages/desktop-main/src/services/FirstRunWizardService.ts
+++ b/app/packages/desktop-main/src/services/FirstRunWizardService.ts
@@ -46,8 +46,18 @@ export class FirstRunWizardService {
     }
   }
 
-  /** Persists `step` so the wizard resumes here if the app is closed and reopened before completion. */
+  /**
+   * Persists `step` so the wizard resumes here if the app is closed and
+   * reopened before completion. `step`'s compile-time type is erased at the
+   * IPC boundary — the caller in `WizardController.saveProgress` is only as
+   * trustworthy as the renderer process — so this validates against
+   * {@link WIZARD_STEPS} before writing, rather than silently persisting an
+   * unsupported value that `getProgress` would later discard anyway.
+   */
   async recordStep(step: WizardStepName): Promise<void> {
+    if (!WIZARD_STEPS.includes(step)) {
+      throw new Error(`Unsupported wizard step: ${String(step)}`);
+    }
     const path = this.stateFilePath();
     await mkdir(dirname(path), { recursive: true });
     await writeFile(path, JSON.stringify({ step } satisfies WizardProgress), 'utf-8');
@@ -61,10 +71,19 @@ export class FirstRunWizardService {
    * "Reconfigure" flow) would call `getProgress()` and jump straight back
    * to whatever step was last recorded (often `terraform-init`), skipping
    * every earlier step with none of their answers rehydrated.
+   *
+   * @remarks
+   * The resume file is removed *before* setting the completion flag, not
+   * after: if `rm` were to throw with the flag already set, the caller
+   * would see `complete()` reject (a failure) while the store already says
+   * the wizard is done — a future launch would then skip the wizard
+   * entirely despite the IPC call having reported failure, with the stale
+   * resume file still lingering. Removing first means any failure here
+   * leaves the wizard genuinely incomplete, matching what the caller was told.
    */
   async complete(): Promise<void> {
-    this.store.set('wizardCompleted', true);
     await rm(this.stateFilePath(), { force: true });
+    this.store.set('wizardCompleted', true);
   }
 
   /** Absolute path to the resumable state file. Extracted as a seam so tests can `vi.spyOn` it. */
--- a/app/packages/desktop-main/src/services/FirstRunWizardService.ts
+++ b/app/packages/desktop-main/src/services/FirstRunWizardService.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@nestjs/common';
-import { mkdir, readFile, writeFile } from 'fs/promises';
+import { mkdir, readFile, rm, writeFile } from 'fs/promises';
 import { join, dirname } from 'path';
 import { tmpdir } from 'os';
 import { createRequire } from 'module';
@@ -56,11 +56,15 @@ export class FirstRunWizardService {
   /**
    * Marks the wizard complete. Setting `wizardCompleted: true` in
    * `ElectronStoreService` is what actually gates the app router past the
-   * wizard (`WizardController.getState`) — the resume file is left in place
-   * but becomes irrelevant once this flag is set.
+   * wizard (`WizardController.getState`). Also clears the resume file —
+   * without this, a future re-entry into the wizard (e.g. #211's Settings
+   * "Reconfigure" flow) would call `getProgress()` and jump straight back
+   * to whatever step was last recorded (often `terraform-init`), skipping
+   * every earlier step with none of their answers rehydrated.
    */
-  complete(): void {
+  async complete(): Promise<void> {
     this.store.set('wizardCompleted', true);
+    await rm(this.stateFilePath(), { force: true });
   }
 
   /** Absolute path to the resumable state file. Extracted as a seam so tests can `vi.spyOn` it. */
@@ -70,10 +74,17 @@ export class FirstRunWizardService {
 
   /**
    * Resolves a writable per-user directory: the Electron `userData` path
-   * when running inside Electron, or the OS temp directory in
+   * when running inside Electron, or a namespaced OS temp subdirectory in
    * plain-Node/test contexts. Mirrors `ConfigService.readUserDataPath`'s
    * dynamic-require seam so this module has no static `electron` import
    * (which would fail to resolve outside an Electron process).
+   *
+   * @remarks
+   * The tmpdir fallback is namespaced under `hyveon-wizard` (mirroring
+   * `ConfigService.getRunsDir`'s `hyveon-runs` prefix) rather than writing
+   * directly into the bare, world-writable `tmpdir()` root — an unnamespaced
+   * path there is guessable and poses a symlink-pre-creation risk, since
+   * `writeFile` follows symlinks.
    */
   protected userDataPath(): string {
     if (process.versions['electron']) {
@@ -85,6 +96,6 @@ export class FirstRunWizardService {
         // Fall through to the tmpdir fallback below.
       }
     }
-    return tmpdir();
+    return join(tmpdir(), 'hyveon-wizard');
   }
 }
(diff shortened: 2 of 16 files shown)


app/packages/desktop-main/src/services/FirstRunWizardService.ts
```diff
@@ -0,0 +1,97 @@
...
+@Injectable()
+export class FirstRunWizardService {
+  constructor(private readonly store: ElectronStoreService) {}
+
+  /** Reads the last-recorded step, defaulting to `prerequisites` when the file is missing, unreadable, or corrupt. */
+  async getProgress(): Promise<WizardProgress> {
+    try {
+      const raw = await readFile(this.stateFilePath(), 'utf-8');
+      const parsed = JSON.parse(raw) as Partial<WizardProgress>;
+      if (parsed.step && (WIZARD_STEP_NAMES as readonly string[]).includes(parsed.step)) {
```
CoderCoco: why not just type WIZARD_STEP_NAMES as a readonly string[]? What's up with the cast here?

app/packages/desktop-main/src/services/FirstRunWizardService.ts
```diff
@@ -0,0 +1,97 @@
...
+   * when running inside Electron, or the OS temp directory in
+   * plain-Node/test contexts. Mirrors `ConfigService.readUserDataPath`'s
+   * dynamic-require seam so this module has no static `electron` import
+   * (which would fail to resolve outside an Electron process).
+   */
+  protected userDataPath(): string {
+    if (process.versions['electron']) {
+      try {
+        const require = createRequire(import.meta.url);
+        const electron = require('electron') as { app: { getPath(name: string): string } };
```
CoderCoco, line 112: Why the dynamic require instead of importing?

app/packages/web/src/components/first-run-wizard/wizard.utils.ts
```diff
@@ -7,10 +7,9 @@
 import type { PrerequisitesReport } from '@hyveon/desktop-preload';
 
 /**
- * Ordered wizard steps. Later PRs in this epic (#210 terraform-init) append
- * to this array.
+ * Ordered wizard steps.
  */
-export const WIZARD_STEPS = ['prerequisites', 'pick-cloud', 'credentials', 'bootstrap'] as const;
+export const WIZARD_STEPS = ['prerequisites', 'pick-cloud', 'credentials', 'bootstrap', 'terraform-init'] as const;
```
CoderCoco: Is this duplicated in an earlier file?

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
```diff
@@ -76,6 +84,45 @@ export function FirstRunWizard() {
 
   const step = WIZARD_STEPS[stepIndex];
 
+  // Resume-on-mount: jump straight to the last-recorded step instead of
+  // always restarting at `prerequisites`, so closing and reopening the app
+  // mid-flow doesn't lose progress. Any failure (including a missing IPC
+  // bridge) leaves `stepIndex` at its default of 0.
+  useEffect(() => {
```
CoderCoco, line 107: **Resume restores the step, but not the answers behind it.**

Every answer this wizard collects lives in local `useState` in this component, seeded from defaults: `resourceNames` from `defaultBootstrapResourceNames()`, `region`/`pasteRegion` as `''`, `credentialMode` as `'profile'`, `selectedProfileName` as `''`. Nothing here rehydrates them from `ElectronStoreService` before the jump.

Concrete failure: operator renames the state bucket to `acme-tfstate` on the bootstrap step, advances to terraform-init, closes the app. On relaunch `getProgress()` returns `terraform-init`, `setStepIndex(4)` fires, `TerraformInitStep` mounts, and its mount effect *immediately* runs

```
terraform init -backend-config=bucket=hyveon-tfstate -backend-config=region= -backend-config=dynamodb_table=hyveon-tflock
```

against a backend the operator never asked for — no confirmation, no visible indication the values were reset. Resuming to `bootstrap` is milder but still silently drops custom resource names and shows all three rows back at `pending`.

Two ways out, either is fine:

1. Clamp the resume to the last step whose preconditions are actually reconstructible — e.g. never resume past `credentials` unless the answers were rehydrated first.
2. Rehydrate before honouring the jump: `wizard.state.get` already returns `{ activeCloud, aws: { profile, region } }`, and the bootstrap resource names would need to join it (either persisted in `wizard-state.json` alongside `step`, or in the store).

Given `WizardProgress` is `{ step }` today, widening it to carry the bootstrap resource names is probably the smallest honest fix — `getProgress`'s existing "unrecognized shape → `DEFAULT_PROGRESS`" guard already gives you a safe migration path for old files.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
```diff
@@ -76,6 +84,45 @@ export function FirstRunWizard() {
...
+  // shouldn't block the wizard, only degrade resume on the next launch.
+  useEffect(() => {
+    if (!window.gsd) return;
+    window.gsd.wizard.saveProgress({ step }).catch(() => {});
+  }, [step]);
+
+  const backendConfig = useMemo<TerraformInitConfig>(
+    () => ({
+      bucket: resourceNames.stateBucket,
+      region: credentialMode === 'profile' ? region : pasteRegion,
```
CoderCoco, line 144: `region` can legitimately be `''` here even on the straight-through path — nothing gates advancement on it.

`advanceDisabled` for the credentials step is `!credentialsChosen`, and `credentialsChosen` only checks `selectedProfileName !== ''` (profile mode) or `pastedProfileName !== null` (paste mode). `selectProfile()` sets `setRegion(profile.region ?? '')`, so picking a profile with no `region` in its `~/.aws/config` block leaves `region` empty and Next stays enabled. Same in paste mode — `submitPaste` sends `region: pasteRegion || undefined`, so an empty paste region is explicitly allowed through.

That lands as `-backend-config=region=` on the init command line. The failure is at least visible (the step goes to `failed` and offers Retry), but the message the operator sees will be a raw terraform backend error rather than "pick a region."

Either gate the credentials step on a non-empty region, or have `TerraformInitStep` refuse to start and say so when `backendConfig.region` is blank.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
```diff
@@ -76,6 +84,45 @@ export function FirstRunWizard() {
...
+      });
+  }, []);
+
+  // Persists the current step on every change (including the resume-on-mount
+  // jump above) so `userData/wizard-state.json` stays in sync with what the
+  // operator is actually looking at. Fire-and-forget: a failed write here
+  // shouldn't block the wizard, only degrade resume on the next launch.
+  useEffect(() => {
+    if (!window.gsd) return;
+    window.gsd.wizard.saveProgress({ step }).catch(() => {});
```
CoderCoco, line 138: This effect fires on mount with `step === 'prerequisites'` and writes it, while the resume effect above is concurrently reading the same file.

Today the read wins because effects run in declaration order and `readFile` is dispatched before `writeFile` — but that's the only thing keeping it correct, and it isn't stated anywhere. If the write ever lands first, the recorded step is clobbered to `prerequisites` and resume is silently gone.

Cheapest fix: gate the save on the resume having settled.

```ts
const resumed = useRef(false);
// in the resume effect's .then()/.catch(): resumed.current = true;
useEffect(() => {
  if (!window.gsd || !resumed.current) return;
  window.gsd.wizard.saveProgress({ step }).catch(() => {});
}, [step]);
```

(That also stops the redundant no-op write of `prerequisites` on every single launch.)

app/packages/desktop-main/src/services/FirstRunWizardService.ts
```diff
@@ -0,0 +1,97 @@
...
+        return electron.app.getPath('userData');
+      } catch {
+        // Fall through to the tmpdir fallback below.
+      }
+    }
+    return tmpdir();
```
CoderCoco: This resolves to a bare `/tmp/wizard-state.json`, unlike the seam it says it mirrors — `ConfigService.getRunsDir()` namespaces its equivalent fallback as `join(tmpdir(), 'hyveon-runs')`.

Two consequences on Linux/macOS, where `/tmp` is world-writable:

- Any other user on the box can pre-create `/tmp/wizard-state.json` as a symlink to a file owned by the app user; `writeFile` follows symlinks, so `recordStep()` becomes an arbitrary-file write as that user. The sticky bit stops deletion, not pre-creation.
- Collides across users on a shared machine.

`join(tmpdir(), 'hyveon', 'wizard-state.json')` (or reuse whatever `hyveon-runs` prefix you prefer) closes both. `recordStep`'s `mkdir(..., { recursive: true })` already creates the parent, so this is a one-line change.

app/packages/desktop-main/src/services/FirstRunWizardService.ts
```diff
@@ -0,0 +1,97 @@
...
+    await writeFile(path, JSON.stringify({ step } satisfies WizardProgress), 'utf-8');
+  }
+
+  /**
+   * Marks the wizard complete. Setting `wizardCompleted: true` in
+   * `ElectronStoreService` is what actually gates the app router past the
+   * wizard (`WizardController.getState`) — the resume file is left in place
+   * but becomes irrelevant once this flag is set.
+   */
+  complete(): void {
```
CoderCoco: The TSDoc's "the resume file is left in place but becomes irrelevant once this flag is set" holds only as long as nothing re-enters the wizard.

Task 12 in the same `tasks.md` — issue #211, "Reconfigure entry point in Settings" — is exactly that. Once it lands, re-entering the wizard will hit `getProgress()` → `terraform-init` and jump straight to the last step, skipping prerequisites/cloud/credentials/bootstrap entirely, and (per the resume comment above) do it with all answers reset to defaults.

Suggest `complete()` also clears the file (`rm(this.stateFilePath(), { force: true })`, or `recordStep('prerequisites')`) so the next entry starts clean. It'd make this method async, which the controller already tolerates.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
```diff
@@ -9,13 +9,14 @@
  * Prerequisites, pick-cloud, credentials, and bootstrap exist so far; a
  * later PR in this epic (#210) appends the terraform-init step.
```
CoderCoco: Stale — this *is* the PR that appends the terraform-init step. Worth rewording to list all five steps, since the file-header TSDoc is the first thing anyone reads here.

CoderCoco (commented):
Honest review — the step component, the IPC surface, and the test coverage are all in good shape (the `getProgress` degrade-never-throw contract, the streamId-aware `terraform.init` consumption, and the RTL specs are genuinely thorough). Two things I think are real bugs, and a few smaller ones.

**The one that matters:** resume-on-mount restores the *step index* but none of the *answers*. Every answer in this shell (`resourceNames`, `region`, `pasteRegion`, `credentialMode`, `selectedProfileName`) is local `useState` seeded from defaults, and nothing rehydrates them. Resuming to `terraform-init` therefore auto-runs `terraform init` against `defaultBootstrapResourceNames()` and `region: ''` — i.e. against a backend the operator never configured. Details inline.

**Summary of findings**

| # | Where | What |
|---|-------|------|
| 1 | `first-run-wizard.component.tsx:91` | Resume jumps the step index without restoring answers → `terraform init` runs against default bucket + empty region |
| 2 | `first-run-wizard.component.tsx:113` | `backendConfig.region` can be `''` on the happy path too — nothing gates Next on a non-empty region |
| 3 | `first-run-wizard.component.tsx:110` | Save-on-mount races the resume read; correctness rests on IPC dispatch ordering |
| 4 | `FirstRunWizardService.ts:95` | Bare `tmpdir()` (not namespaced like `ConfigService.getRunsDir()`) → `/tmp/wizard-state.json` |
| 5 | `FirstRunWizardService.ts:69` | `complete()` leaves the resume file pinned at `terraform-init`, which #211 (next task) will trip over |
| 6 | `first-run-wizard.component.tsx:10` | Stale TSDoc — says #210 is a later PR |
| 7 | `FirstRunWizardService.ts:14` | Step-name list now triplicated with three "keep in sync" comments |

1 and 2 are the ones I'd want addressed before merge; the rest are cheap.

Test plan claims (2080 passing, lint clean, build clean) — I haven't re-run them, taking those as stated.

app/packages/desktop-main/src/services/FirstRunWizardService.ts
```diff
@@ -0,0 +1,97 @@
...
+import { createRequire } from 'module';
+import { ElectronStoreService } from './ElectronStoreService.js';
+
+/**
+ * Ordered first-run wizard steps whose progress is resumable via
+ * `userData/wizard-state.json`. Mirrors `WIZARD_STEPS` in
+ * `@hyveon/web`'s `wizard.utils.ts` — that file is the renderer's source of
+ * truth for step ordering; keep this list in sync with it.
+ */
+export const WIZARD_STEP_NAMES = ['prerequisites', 'pick-cloud', 'credentials', 'bootstrap', 'terraform-init'] as const;
```
CoderCoco: This is now the third copy of the step list, each carrying its own "keep in sync" comment:

- `WIZARD_STEP_NAMES` here (runtime validation — the one that actually rejects unknown values)
- `WizardStepName` in `desktop-preload/src/gsd-api.ts:419` (type only)
- `WIZARD_STEPS` in `web/.../wizard.utils.ts:12` (runtime, drives ordering)

The first and third have to agree *by value*, not just by type, or a step recorded by the renderer gets silently rejected by `getProgress()` and the operator lands back at step 1 with no error. A comment can't enforce that.

`@hyveon/shared` already exists for exactly this (`canRun`, command descriptors) and is imported by both server and Lambdas — hoisting the const there and re-exporting would make the drift a compile error instead of a runtime regression. Not blocking, but the comment-based approach is one step away from breaking.

app/packages/desktop-main/src/services/FirstRunWizardService.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  let store: ElectronStoreService;
+
+  beforeEach(() => {
+    scratchDir = mkdtempSync(join(tmpdir(), 'wizard-state-test-'));
+    statePath = join(scratchDir, 'nested', 'wizard-state.json');
+    mkdirSync(join(scratchDir, 'nested'));
+    store = makeStore();
+    service = new TestableFirstRunWizardService(store);
+    vi.spyOn(service, 'stateFilePath').mockReturnValue(statePath);
+  });
```
CoderCoco, line 38: Nit on coverage, not correctness: because every test stubs `stateFilePath` wholesale, `userDataPath()` has no coverage at all — neither the `process.versions['electron']` branch nor the `tmpdir()` fallback nor the `catch` that falls through. That's the method most likely to be wrong in a packaged build (and the one flagged for the `/tmp` namespacing issue). A small test that spies `userDataPath` and asserts `stateFilePath()` composes `<dir>/wizard-state.json` would at least pin the filename and the join.

app/packages/desktop-main/src/services/FirstRunWizardService.ts
CoderCoco: Fixed in e6490bb — removed the cast. It turned out to be unnecessary: `parsed.step` is already typed `WizardStepName | undefined` (from the `as Partial<WizardProgress>` assertion), and once narrowed by the preceding `parsed.step &&` truthy check, its type already matches `WIZARD_STEPS`'s literal-union element type, so `WIZARD_STEPS.includes(parsed.step)` type-checks with no cast at all. Verified in isolation with a standalone tsc check before removing it.

Separately, also hoisted the step list itself into `@hyveon/shared/wizardSteps.ts` (see the reply on the wizard.utils.ts duplication thread) — the local `WIZARD_STEP_NAMES` const is gone, replaced by an import of the shared `WIZARD_STEPS`.

app/packages/desktop-main/src/services/FirstRunWizardService.ts
CoderCoco: Declined — intentional, matches an existing established pattern. This mirrors `ConfigService.readUserDataPath()` exactly (see that method's TSDoc), which uses the same `createRequire` + lazy `require('electron')` seam for the same reason: under plain Node (e.g. this vitest process, or any test/build context), `require("electron")" resolves to a string (the path to the Electron binary) rather than the real module object — Electron's module-resolution shim only substitutes the real API when running inside an actual Electron process. A static `import { app } from "electron"` at the top of this file would try to destructure `.app" off that string and throw at *import time*, before the `process.versions["electron"]" runtime guard ever gets a chance to run. The dynamic require, gated behind that guard, defers touching the module until we already know we are inside Electron. Keeping this file consistent with `ConfigService"'s established convention rather than introducing a second pattern for the same problem.

app/packages/web/src/components/first-run-wizard/wizard.utils.ts
CoderCoco: Confirmed and fixed in e6490bb — moved the ordered step list into `@hyveon/shared/wizardSteps.ts` (`WIZARD_STEPS` + `WizardStep` type), mirroring the `MINIMUM_TERRAFORM_VERSION`/`GAME_SERVER_DEPLOY_ALL_ACTIONS` precedent already in that package. Both this file and `FirstRunWizardService.ts` (desktop-main) now import from there instead of each declaring their own hand-synced copy; `wizard.utils.ts` re-exports `WIZARD_STEPS`/`WizardStep` so no call sites needed to change their import path.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
CoderCoco: Fixed in 7a05c49 — went with option 1 (clamp), not full rehydration. Resume now clamps to at most the `bootstrap` step: `Math.min(index, WIZARD_STEPS.indexOf('bootstrap'))`. Bootstrap is safe to resume into as-is — its IPC calls (`bootstrapStateBucket` etc.) read region from the already-persisted `ElectronStoreService.aws.region` (written by `wizard.state.save` when leaving the credentials step), not from any frontend-only state — so the worst case there is the operator re-clicking "Bootstrap AWS resources" with default names. terraform-init is the one that's unsafe to auto-resume into (backendConfig.region has no such backend fallback), so that's exactly what the clamp excludes. Added a regression test asserting a recorded terraform-init step lands on the bootstrap step's UI and gsd.terraform.init is never called.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
CoderCoco: Fixed in 7a05c49 — gated the credentials step's `credentialsChosen` (and therefore Next) on a non-empty region in both modes: `region !== ''` for profile mode, `pasteRegion !== ''` for paste mode. Added tests: selecting a profile with no configured region (`personal` in the test fixtures) now keeps Next disabled until the operator fills in a region, and the paste-mode 'no region entered' case is covered the same way.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
CoderCoco: Fixed in 7a05c49, using your suggested pattern. Added a `resumeSettledRef`, set synchronously inside the resume promise's own `.then()`/`.catch()` (not a trailing `.finally()` — I tried that first and it turned out not to be early enough; a test asserting saveProgress hadn't fired before resume settled caught it), and the save effect bails out until that ref is true. Also confirmed the side benefit you called out: the redundant no-op `prerequisites` write on every launch is gone now, since `setStepIndex` (and therefore a `step` change that would re-trigger the save effect) is only called when `index > 0`.

app/packages/desktop-main/src/services/FirstRunWizardService.ts
CoderCoco: Fixed in 7a05c49 — `userDataPath()`'s tmpdir fallback now returns `join(tmpdir(), 'hyveon-wizard')` instead of the bare `tmpdir()` root, matching `ConfigService.getRunsDir()`'s `hyveon-runs` prefix. Added a test pinning the exact fallback path.

app/packages/desktop-main/src/services/FirstRunWizardService.ts
CoderCoco: Fixed in 7a05c49 — `complete()` is now `async` and clears the resume file (`rm(this.stateFilePath(), { force: true })`) after setting `wizardCompleted`. Updated `WizardController.complete()` to `await` it. Added a test asserting a recorded step is gone (`getProgress()` back to `prerequisites`) after `complete()` runs, plus a not-founds-anything-to-delete case.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
CoderCoco: Fixed in 7a05c49 — reworded to list all five steps instead of the stale "a later PR in this epic (#210) appends the terraform-init step" line.

app/packages/desktop-main/src/services/FirstRunWizardService.ts
CoderCoco: Same finding as a comment left directly on this PR — fixed in e6490bb by hoisting the step list into `@hyveon/shared/wizardSteps.ts` (`WIZARD_STEPS` + `WizardStep`), imported by both this file and `wizard.utils.ts` instead of each declaring its own copy. `desktop-preload/gsd-api.ts`'s `WizardStepName` stays a type-only mirror, consistent with how every other cross-package type in that file is handled (it never imports from `@hyveon/shared`, by established convention) — only the two *runtime* arrays needed to be a compile-time-enforced single source of truth.

app/packages/desktop-main/src/services/FirstRunWizardService.test.ts
CoderCoco: Fixed in 7a05c49 — added two tests: one spies `userDataPath` directly and asserts `stateFilePath()` composes `<dir>/wizard-state.json`, the other exercises the real (unspied) `userDataPath()` in this Node test context and asserts the `hyveon-wizard`-namespaced tmpdir fallback (addressing the sibling thread on the symlink risk) is what's actually returned. Didn't add electron-branch coverage — mocking `process.versions.electron` plus a fake `electron` module for a one-line `getPath` passthrough felt like more test complexity than the branch warrants; `ConfigService`'s equivalent branch doesn't have dedicated coverage either, for the same reason.

CoderCoco:
## Fixes Applied Successfully

Fixed 3 file(s) based on 3 CodeRabbit feedback item(s).

**Files modified:**
- `app/packages/desktop-main/src/services/FirstRunWizardService.ts`
- `app/packages/desktop-main/src/services/FirstRunWizardService.test.ts`
- `app/packages/desktop-main/src/controllers/wizard.controller.test.ts`

**Changes:**
- `complete()` now removes the resume file before setting `wizardCompleted`, not after — a failed `rm()` no longer leaves the store saying "complete" while the resume file lingers.
- `recordStep()` now validates `step` against `WIZARD_STEPS` and rejects unsupported values instead of silently persisting something `getProgress()` would later discard anyway.
- Test stub factories (`makeStore`, `makeFirstRunWizard`) now assign to an explicit `Partial<T>` variable before the final `as T` cast, instead of chaining both assertions inline.

**Commit:** `38cb01e`

The latest autofix changes are on the `claude/issue-210-wizard-terraform-init` branch.

app/packages/desktop-main/src/services/FirstRunWizardService.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+  public override userDataPath(): string {
+    return super.userDataPath();
+  }
+}
+
+/** Build an `ElectronStoreService` stub whose `set()` calls are observable. */
+function makeStore(): ElectronStoreService {
+  return { set: vi.fn() } as Partial<ElectronStoreService> as ElectronStoreService;
+}
```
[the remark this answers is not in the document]
CoderCoco: Fixed in 38cb01e — both `makeStore` (this file) and `makeFirstRunWizard` (wizard.controller.test.ts) now assign to an explicit `Partial<T>` variable before the final `as T` cast, instead of chaining both assertions inline.

app/packages/desktop-main/src/services/FirstRunWizardService.ts
```diff
@@ -0,0 +1,101 @@
...
+    } catch {
+      return DEFAULT_PROGRESS;
+    }
+  }
+
+  /** Persists `step` so the wizard resumes here if the app is closed and reopened before completion. */
+  async recordStep(step: WizardStepName): Promise<void> {
+    const path = this.stateFilePath();
+    await mkdir(dirname(path), { recursive: true });
+    await writeFile(path, JSON.stringify({ step } satisfies WizardProgress), 'utf-8');
```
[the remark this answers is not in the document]
CoderCoco: Fixed in 38cb01e — `recordStep()` now throws `Unsupported wizard step: ...` when `step` isn't in `WIZARD_STEPS`, before ever touching the filesystem. Added a regression test that bypasses the compile-time type to exercise the runtime guard and asserts nothing gets written.

app/packages/desktop-main/src/services/FirstRunWizardService.ts
```diff
@@ -0,0 +1,101 @@
...
+   * `ElectronStoreService` is what actually gates the app router past the
+   * wizard (`WizardController.getState`). Also clears the resume file —
+   * without this, a future re-entry into the wizard (e.g. #211's Settings
+   * "Reconfigure" flow) would call `getProgress()` and jump straight back
+   * to whatever step was last recorded (often `terraform-init`), skipping
+   * every earlier step with none of their answers rehydrated.
+   */
+  async complete(): Promise<void> {
+    this.store.set('wizardCompleted', true);
+    await rm(this.stateFilePath(), { force: true });
```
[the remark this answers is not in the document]
CoderCoco: Fixed in 38cb01e — flipped the order so `rm()` runs before `this.store.set('wizardCompleted', true)`. A failed removal now leaves the wizard genuinely incomplete (matching what the caller is told) instead of the store already saying "done" with a stale resume file still on disk.

PR #331, opened 2026-07-26
feat(web): Reconfigure entry point in Settings

### LINKED ISSUE #139: epic: first-run wizard + credentials UX
## Epic E: first-run wizard + credentials UX

> Part of the **Electron desktop pivot**. Authoritative scope and rationale live in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Scope

Multi-step wizard that runs on first launch and is re-runnable from Settings. Steps: detect prerequisites (`terraform`, `aws`); pick cloud (v1: AWS only); pick or paste credentials; bootstrap the S3 state bucket, DynamoDB lock table, and S3 tfvars bucket via AWS SDK calls (no shell-out); run `terraform init`. Pasted credentials encrypted via `safeStorage`.

## Children

- [ ] E.1 — #182 feat(desktop-main): prerequisite-detection service (terraform, aws on PATH)
- [ ] E.2 — #184 feat(web): wizard step — install prerequisites
- [ ] E.3 — #186 feat(web): wizard step — pick cloud (v1: AWS only)
- [ ] E.4 — #189 feat(desktop-main): AwsProfileService — list and read ~/.aws/credentials
- [ ] E.5 — #192 feat(web): wizard step — pick or paste credentials
- [ ] E.6 — #197 feat(desktop-main): safeStorage paste-flow for credentials
- [ ] E.7 — #200 feat(desktop-main): SDK bootstrap of S3 state bucket
- [ ] E.8 — #203 feat(desktop-main): SDK bootstrap of DynamoDB lock table
- [ ] E.9 — #205 feat(desktop-main): SDK bootstrap of S3 tfvars bucket
- [ ] E.10 — #208 feat(desktop-main): IAM SimulatePrincipalPolicy check against GameServerDeployAll
- [ ] E.11 — #210 feat(web): wizard step — terraform init with live log
- [ ] E.12 — #211 feat(web): Reconfigure entry point in Settings

## Acceptance
- Fresh install on a clean machine reaches "dashboard ready" through the wizard without manual file editing.
- Pasted access keys never appear in plaintext in `electron-store.json` (verified via integration test).
- `iam:SimulatePrincipalPolicy` flags any missing IAM actions against `GameServerDeployAll` and surfaces a copy-paste-able JSON.
- Wizard answers are reloaded correctly on the next launch.
- "Reconfigure" entry point in Settings reruns the wizard against the current install.

### LINKED ISSUE #211: feat(web): Reconfigure entry point in Settings

CoderCoco:
Closes #211
Closes #139

## Summary

- `FirstRunWizard` gains a `mode: 'first-run' | 'reconfigure'` prop. In `'reconfigure'` mode it skips the `prerequisites` step, prefills `pick-cloud`/`credentials` from the durably-stored `wizard.state.get`, and pre-marks `pick-cloud`/`credentials`/`bootstrap` as completed with a per-step "Edit" affordance (reaching `terraform-init` always re-runs it, since that's the explicit re-run the operator asked for by navigating there).
- Edits made while reconfiguring are buffered in local component state rather than persisted per-step — a single `wizard.state.save` call (via a new `onBeforeFinish` hook on `TerraformInitStep`) commits everything right before `wizard.complete` runs on the Finish click, so a mid-flow Cancel writes nothing and the pre-existing configuration is untouched.
- Settings gets a new "Cloud Setup" section: the Terraform version resolved by `wizard.prereqs.check` alongside the pinned `MINIMUM_TERRAFORM_VERSION`, and a "Reconfigure" button that swaps the page body for the wizard shell in `mode="reconfigure"`.
- This is the last child of epic #139 (first-run wizard + credentials UX) — all 11 sibling issues are already closed.

## Test plan

- [x] `npm run app:test` — 2096/2096 tests pass, including new specs covering: Reconfigure entry renders, completed-step Edit affordances, single-field edit preserving every other stored setting, mid-flow Cancel committing nothing, and the Terraform version display.
- [x] `npm run app:lint` — clean (pre-existing unrelated warnings in `ConfigService.ts` only).
- [x] `npm run app:build` — full `shared → cloud-aws → desktop-main → web` pipeline compiles cleanly.

### DIFF
--- a/app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
+++ b/app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
@@ -11,6 +11,7 @@
  * wizard).
  */
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { CheckCircle2 } from 'lucide-react';
 import type { AwsProfileSummary, IamCheckResult, PrerequisitesReport, TerraformInitConfig } from '@hyveon/desktop-preload';
 import { Button } from '@/components/ui/button.component';
 import { PrerequisitesStep } from './prerequisites-step.component.js';
@@ -22,6 +23,7 @@ import {
   WIZARD_STEPS,
   arePrerequisitesSatisfied,
   defaultBootstrapResourceNames,
+  reconfigureSteps,
   type BootstrapResourceKey,
   type BootstrapResourceState,
   type WizardStep,
@@ -36,10 +38,33 @@ const STEP_LABELS: Record<WizardStep, string> = {
   'terraform-init': 'Finish setup',
 };
 
+/**
+ * Steps in this list start collapsed to a completed summary (with an Edit
+ * affordance) in `mode: 'reconfigure'`, since Settings only offers
+ * Reconfigure once the wizard has already completed once — every one of
+ * these already has a real answer on record. `terraform-init` is excluded:
+ * it has no standalone "answer" to summarize, and reaching it is itself the
+ * explicit re-run the operator asked for by clicking through to it.
+ */
+const RECONFIGURE_PRE_COMPLETED_STEPS: WizardStep[] = ['pick-cloud', 'credentials', 'bootstrap'];
+
 /** Props for {@link FirstRunWizard}. */
 export interface FirstRunWizardProps {
   /** Invoked once the terraform-init step's `wizard.complete` call succeeds. */
   onComplete?: () => void;
+  /**
+   * `'first-run'` (default) gates the whole app and runs all five steps,
+   * persisting `pick-cloud`/`credentials` answers immediately via
+   * `wizard.state.save` as the operator advances. `'reconfigure'` (#211,
+   * launched from Settings) skips `prerequisites`, pre-marks
+   * `pick-cloud`/`credentials`/`bootstrap` as completed with a per-step Edit
+   * affordance, and buffers any edited answers locally — committed in one
+   * `wizard.state.save` call right before `wizard.complete` runs, so a
+   * mid-flow Cancel writes nothing.
+   */
+  mode?: 'first-run' | 'reconfigure';
+  /** `'reconfigure'`-only: invoked when the operator cancels without finishing. Never called in `'first-run'` mode. */
+  onCancel?: () => void;
 }
 
 /**
@@ -49,7 +74,8 @@ export interface FirstRunWizardProps {
  * paste form). Fetches an initial prerequisites check and AWS profile list
  * on mount.
  */
-export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
+export function FirstRunWizard({ onComplete, mode = 'first-run', onCancel }: FirstRunWizardProps = {}) {
+  const steps = useMemo(() => (mode === 'reconfigure' ? reconfigureSteps() : WIZARD_STEPS), [mode]);
   const [stepIndex, setStepIndex] = useState(0);
   const [report, setReport] = useState<PrerequisitesReport | null>(null);
   const [checking, setChecking] = useState(false);
@@ -83,7 +109,14 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
   const [iamChecking, setIamChecking] = useState(false);
   const [iamError, setIamError] = useState<string | null>(null);
 
-  const step = WIZARD_STEPS[stepIndex];
+  // Reconfigure-only: which pre-completed steps are collapsed to a summary
+  // (present in the set) vs. expanded for editing (removed from it). Empty —
+  // and unused — in `'first-run'` mode.
+  const [completedSteps, setCompletedSteps] = useState<Set<WizardStep>>(
+    () => new Set(mode === 'reconfigure' ? RECONFIGURE_PRE_COMPLETED_STEPS : []),
+  );
+
+  const step = steps[stepIndex];
 
   // Guards the save-progress effect below until the resume-on-mount effect
   // has settled (resolved or rejected) — otherwise the two effects race to
@@ -104,7 +137,14 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
   // safe by comparison: its IPC calls read the region from the credentials
   // step's already-persisted `wizard.state.save` call, and worst case the
   // operator just has to re-click "Bootstrap AWS resources".
+  //
+  // Both this effect and the save-progress effect below are `'first-run'`-only:
+  // `userData/wizard-state.json` tracks resumable progress through the
+  // *gating* wizard, which is meaningless for Reconfigure (the app is already
+  // past the gate, and Reconfigure has its own pre-completed-steps/Cancel
+  // model instead of resume).
   useEffect(() => {
+    if (mode !== 'first-run') return;
     if (!window.gsd) {
       resumeSettledRef.current = true;
       return;
@@ -125,7 +165,7 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
         // Best-effort — starting over at step 1 is always a safe fallback.
         resumeSettledRef.current = true;
       });
-  }, []);
+  }, [mode]);
 
   // Persists the current step on every change (including the resume-on-mount
   // jump above) so `userData/wizard-state.json` stays in sync with what the
@@ -134,9 +174,43 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
   // Gated on `resumeSettledRef` so this can never race the resume effect's
   // own read of the same file (see that effect's comment).
   useEffect(() => {
-    if (!window.gsd || !resumeSettledRef.current) return;
+    if (mode !== 'first-run' || !window.gsd || !resumeSettledRef.current) return;
     window.gsd.wizard.saveProgress({ step }).catch(() => {});
-  }, [step]);
+  }, [mode, step]);
+
+  // Reconfigure-only: prefills the pick-cloud/credentials answers from the
+  // durably-stored `wizard.state.get` so the collapsed step summaries (and
+  // any step the operator opens for editing) reflect what's actually
+  // configured, not the first-run defaults. Runs once; if it fails, the
+  // steps simply keep their first-run defaults and the operator can still
+  // fill them in via Edit like any other value.
+  useEffect(() => {
+    if (mode !== 'reconfigure' || !window.gsd) return;
+    window.gsd.wizard
+      .getState()
+      .then((state) => {
+        if (state.activeCloud) setSelectedCloud(state.activeCloud);
+        if (state.aws?.profile) {
+          const knownProfile = profiles?.some((p) => p.profileName === state.aws!.profile);
+          if (knownProfile) {
+            setCredentialMode('profile');
+            setSelectedProfileName(state.aws.profile);
+            setRegion(state.aws.region ?? '');
+          } else {
+            // Not in `~/.aws` — must be a `creds.aws.<profile>` pasted-key
+            // entry from a prior paste-flow save (see `AwsProfileService`).
+            setCredentialMode('paste');
+            setPastedProfileName(state.aws.profile);
+            setPasteRegion(state.aws.region ?? '');
+          }
+        }
+      })
+      .catch(() => {
+        // Best-effort — the step just falls back to first-run defaults.
+      });
+    // Depends on `profiles` too, so this re-runs once the list resolves and
+    // the `knownProfile` check above isn't stuck on `null` from before it loaded.
+  }, [mode, profiles]);
 
   const backendConfig = useMemo<TerraformInitConfig>(
     () => ({
@@ -169,8 +243,11 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
   }, []);
 
   useEffect(() => {
+    // `prerequisites` isn't a Reconfigure step — skip the check entirely
+    // rather than running it for no visible step.
+    if (mode !== 'first-run') return;
     void checkPrereqs();
-  }, [checkPrereqs]);
+  }, [mode, checkPrereqs]);
 
   useEffect(() => {
     async function fetchProfiles() {
@@ -305,8 +382,14 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
       ? selectedProfileName !== '' && region !== ''
       : pastedProfileName !== null && pasteRegion !== '';
 
-  const advanceDisabled =
-    step === 'prerequisites'
+  // A collapsed (completed, not being edited) Reconfigure step already has a
+  // real answer on record — Next should never be gated on this render's
+  // local form state, which for an unopened step may not even be prefilled yet.
+  const stepCollapsed = mode === 'reconfigure' && completedSteps.has(step);
+
+  const advanceDisabled = stepCollapsed
+    ? false
+    : step === 'prerequisites'
       ? !arePrerequisitesSatisfied(report)
       : step === 'credentials'
         ? !credentialsChosen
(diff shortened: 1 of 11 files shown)


CoderCoco (commented):
## Review — Reconfigure entry point (#211 / epic #139)

The shape of this is right: one `mode` prop rather than a forked component, a single `wizard.state.save` commit point wired through `onBeforeFinish`, and `saveState`'s server-side `{...current, ...body.aws}` merge means an edited region genuinely does keep the stored profile. The two `'first-run'`-only guards on the resume/save-progress effects are the correct call — Reconfigure has no business writing `wizard-state.json`.

The headline problem is that Reconfigure's **last step runs `terraform init` against backend resource names that were never rehydrated**. Nothing in the app persists the bootstrap step's bucket/table names (`AppStoreSchema` has only `wizardCompleted`, `activeCloud`, `aws`, `creds`), so `backendConfig` in Reconfigure is always `defaultBootstrapResourceNames()` — `hyveon-tfstate` / `hyveon-tflock`. Any operator who edited those names during first run (the step makes them editable on purpose) gets a failing or wrong `terraform init` at the end of every Reconfigure, and the bootstrap step is collapsed by default so they never see the names that are about to be used.

### Findings

| # | Finding | File | Rank |
|---|---------|------|------|
| 1 | `terraform init` in Reconfigure always uses default bootstrap resource names | `first-run-wizard.component.tsx:536` | **fix before merge** |
| 2 | `commitReconfigureAnswers` writes `profile`/`region` unconditionally — a failed prefill clobbers the stored values | `first-run-wizard.component.tsx:462` | **fix before merge** |
| 3 | Prefill effect has no once-guard: transient paste-mode misclassification, and can overwrite in-progress edits | `first-run-wizard.component.tsx:194` | cheap |
| 4 | Cancel button renders even when `onCancel` is undefined (dead control) | `first-run-wizard.component.tsx:565` | cheap |
| 5 | "mid-flow Cancel writes nothing" overstates the guarantee | `first-run-wizard.component.tsx:63` | doc-only |
| 6 | Settings re-runs `checkPrereqs` (two child processes) on every visit to `/settings` | `settings.page.tsx:27` | optional |

One more nit not worth a thread: the `afterEach(cleanup)` added to `settings.page.test.tsx` is redundant — `app/vitest.setup.ts:13` already runs `cleanup()` after each test.

### What I did and did not verify

- **Read statically**: the full wizard component and `terraform-init-step.component.tsx` at head, `wizard.utils.ts`, `settings.page.tsx`, `WizardController.saveState`/`getState`, `AppStoreSchema`, `TerraformService.init`'s `-backend-config` argv, and `BootstrapService`'s per-call credential resolution (which does pick up a changed profile/region without a restart — no finding there).
- **Did not run**: `npm run app:test`, `app:lint`, or `app:build`. Taking the 2096/2096, clean-lint and clean-build claims in the PR body as stated.
- **Did not look at**: the Electron e2e specs, or whether any spec exercises `mode="reconfigure"` outside the new jsdom tests.
- Findings 1 and 2 are read off the code paths directly; finding 3's edit-clobber half depends on an effect ordering I reasoned about rather than observed.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
```diff
@@ -355,86 +440,133 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
...
+              {saveError && (
+                <p role="alert" className="text-sm text-[var(--color-red)]">
+                  {saveError}
+                </p>
+              )}
+            </>
+          ))}
+        {step === 'bootstrap' &&
+          (stepCollapsed ? (
+            <CompletedStepSummary label={STEP_LABELS['bootstrap']} onEdit={() => startEdit('bootstrap')} />
```
CoderCoco, line 587: **Bug: Reconfigure runs `terraform init` against the default backend resource names, not the operator's.**

`resourceNames` is initialised from `defaultBootstrapResourceNames()` and is never rehydrated in Reconfigure — the prefill effect above only restores `activeCloud` and `aws`, and `AppStoreSchema` (`ElectronStoreService.ts:43-63`) has nowhere to store the bucket/table names in the first place. So `backendConfig` (line 215) is always `{ bucket: 'hyveon-tfstate', dynamodbTable: 'hyveon-tflock' }` here, and `TerraformInitStep` fires that at `terraform.init` **on mount** (its effect runs immediately, no confirm step), which shells out to `terraform init -input=false -backend-config=bucket=… -backend-config=dynamodb_table=…` (`TerraformService.ts:789-797`).

Concrete failure: operator's first run set the state bucket to `acme-tfstate` (the bootstrap step makes these fields editable by design) → later they open Settings → Reconfigure → the bootstrap step is collapsed, so they never see a resource name → Next → Next → the final step immediately runs `terraform init` with `bucket=hyveon-tfstate`. With `-input=false` and a `.terraform` dir already pinned to `acme-tfstate`, terraform errors out on the changed backend; there is no `-reconfigure`/`-migrate-state` flag on this path. Their only recovery is Back → Edit bootstrap → retype the exact names — and doing so un-collapses the step, which then gates Next on `bootstrapComplete`, forcing a re-run of all three bootstrap calls.

Suggested fix: persist the bootstrap resource names alongside `aws` in `saveState` (first-run writes them, Reconfigure rehydrates them), and until that exists, don't pre-collapse `bootstrap` in Reconfigure — showing the names the init is about to use is the difference between an operator confirming them and discovering them in a stack trace.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
```diff
@@ -355,86 +440,133 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
...
+   * as `onBeforeFinish` so it runs right before `wizard.complete` on the
+   * Finish click — the single commit point for this whole flow.
+   */
+  async function commitReconfigureAnswers() {
+    if (!window.gsd) {
+      throw new Error('IPC bridge (window.gsd) is not available in this context.');
+    }
+    const profile = credentialMode === 'profile' ? selectedProfileName : (pastedProfileName ?? undefined);
+    const chosenRegion = credentialMode === 'profile' ? region : pasteRegion;
+    await window.gsd.wizard.saveState({
```
CoderCoco: **Bug: this commits `profile`/`region` even when the prefill never produced them, wiping the stored values.**

The prefill effect at line 187 swallows a `getState()` rejection and comments that "the step just falls back to first-run defaults" — but those defaults are exactly what this function then commits. With `credentialMode` at its `'profile'` default, `profile` is `''` and `chosenRegion || undefined` is `undefined`. `WizardController.saveState` merges with `{ ...current, ...body.aws }` (`wizard.controller.ts:131-132`), and an explicit `region: undefined` in the spread overwrites the stored value rather than leaving it alone.

Concrete failure: `wizard.state.get` rejects (or returns an `aws` without `profile`) → operator leaves all three steps collapsed, clicks through, `terraform init` succeeds, clicks Finish → stored `aws` goes from `{ profile: 'acme', region: 'eu-west-1' }` to `{ profile: '', region: undefined }`. Everything downstream that reads it (`BootstrapService.ts:233,309`, `IamCheckService.ts:154`) loses the profile and region, and the operator did not touch the credentials step. The same class of gap lets a blank region reach `terraform init` as `-backend-config=region=`: the `credentialsChosen` guard that exists specifically to prevent that (see the comment at line ~410) is bypassed by `stepCollapsed ? false : …`.

Suggested fix: track which steps were actually opened/edited and only include those keys in the payload — a collapsed step should contribute nothing to the commit. Failing that, refuse to commit (and to advance past `credentials`) until the prefill has resolved with a non-empty profile and region.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
```diff
@@ -134,9 +174,43 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
...
+  // steps simply keep their first-run defaults and the operator can still
+  // fill them in via Edit like any other value.
+  useEffect(() => {
+    if (mode !== 'reconfigure' || !window.gsd) return;
+    window.gsd.wizard
+      .getState()
+      .then((state) => {
+        if (state.activeCloud) setSelectedCloud(state.activeCloud);
+        if (state.aws?.profile) {
+          const knownProfile = profiles?.some((p) => p.profileName === state.aws!.profile);
```
CoderCoco, line 218: **The prefill effect has no once-guard, so it both mislabels a real profile and can revert in-progress edits.**

`profiles` starts as `null`, so the *first* pass through this effect always takes the `else` branch: `profiles?.some(...)` is `undefined`, `knownProfile` is falsy, and the credentials step is switched to `'paste'` with the operator's real `~/.aws` profile name stuffed into `pastedProfileName`. The `[mode, profiles]` dependency corrects that once `listAwsProfiles` resolves, but it does so by issuing a *second* `getState()` and re-applying stored values over whatever is in the form.

Concrete failures:
1. If `listAwsProfiles()` rejects (`profilesError` set, `profiles` stays `null` forever), the operator clicking Edit on credentials sees the paste-keys form. Typing in its region field calls `pasteFieldChange`, which sets `pastedProfileName = null` → `credentialsChosen` is false → Next is disabled. To change only their region they now have to re-paste a full access key/secret pair.
2. Operator clicks Reconfigure → Next → Edit → types `eu-west-1` while `listAwsProfiles` is still in flight → the list resolves, the effect re-runs, and `setRegion(state.aws.region)` silently reverts what they typed. This one is ordering-dependent — I reasoned it out of the dep array rather than observing it, and the window is small — but nothing in the code prevents it.

Suggested fix: gate the whole effect on a `prefilledRef` so it applies stored values exactly once, and wait for `profiles !== null || profilesError` before deciding the profile-vs-paste branch.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
```diff
@@ -355,86 +440,133 @@ export function FirstRunWizard({ onComplete }: FirstRunWizardProps = {}) {
...
         <div className="flex justify-between">
-          <Button type="button" variant="outline" onClick={goBack} disabled={stepIndex === 0 || saving}>
-            Back
-          </Button>
+          <div className="flex gap-2">
+            <Button type="button" variant="outline" onClick={goBack} disabled={stepIndex === 0 || saving}>
+              Back
+            </Button>
+            {mode === 'reconfigure' && (
+              <Button type="button" variant="ghost" onClick={onCancel}>
```
CoderCoco, line 616: **Dead control when `onCancel` is omitted.**

`onCancel` is optional on `FirstRunWizardProps`, but this button renders for every `mode === 'reconfigure'` render regardless. Today `settings.page.tsx` is the only caller and it does pass one, so nothing is broken — but the next caller that renders `<FirstRunWizard mode="reconfigure" />` without it ships a Cancel button that does nothing when clicked, with no type error to catch it.

Suggested fix: `{mode === 'reconfigure' && onCancel && (…)}`, or make `onCancel` required whenever `mode` is `'reconfigure'` via a discriminated-union prop type.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
```diff
@@ -36,10 +38,33 @@ const STEP_LABELS: Record<WizardStep, string> = {
...
   onComplete?: () => void;
+  /**
+   * `'first-run'` (default) gates the whole app and runs all five steps,
+   * persisting `pick-cloud`/`credentials` answers immediately via
+   * `wizard.state.save` as the operator advances. `'reconfigure'` (#211,
+   * launched from Settings) skips `prerequisites`, pre-marks
+   * `pick-cloud`/`credentials`/`bootstrap` as completed with a per-step Edit
+   * affordance, and buffers any edited answers locally — committed in one
+   * `wizard.state.save` call right before `wizard.complete` runs, so a
+   * mid-flow Cancel writes nothing.
```
CoderCoco: **Doc accuracy: "a mid-flow Cancel writes nothing" is narrower than it reads.**

What's true is that no `wizard.state.save` happens before Finish. Two other writes are unconditional and mode-independent:

- `submitPaste()` calls `wizard.saveCredentials` the moment the operator clicks Save in the paste form, which writes encrypted key material into `creds.aws.<profile>`. Cancelling afterwards leaves those keys on disk.
- `runBootstrap()` creates real S3 buckets and a DynamoDB table via the three bootstrap IPC calls. Cancelling after that doesn't remove them.

Both are defensible (the credentials-step spec explicitly wants the paste flow to be immediate, and bootstrap is idempotent), but an operator reading "Cancel writes nothing" would be surprised. Suggested fix: reword to "a mid-flow Cancel leaves the stored wizard answers untouched" and note the two side-effecting steps that are exceptions.

app/packages/web/src/pages/settings.page.tsx
```diff
@@ -1,12 +1,44 @@
...
+ * just flip `reconfiguring` back to `false`; nothing about the surrounding
+ * page (route, layout) needs to change either way.
  */
 export function SettingsPage() {
+  const [reconfiguring, setReconfiguring] = useState(false);
+  const [prereqs, setPrereqs] = useState<PrerequisitesReport | null>(null);
+
+  useEffect(() => {
+    if (!window.gsd?.wizard) return;
+    window.gsd.wizard.checkPrereqs().then(setPrereqs).catch(() => {
```
CoderCoco, line 27: **Optional: this spawns two child processes every time `/settings` is mounted.**

`wizard.prereqs.check` runs `PrerequisiteService.check()`, which shells out for `terraform -version` and `aws --version`. This effect has an empty dep array but no module-level cache, so every navigation to Settings — not just the first — pays for both spawns, purely to render one line of text that changes about as often as the operator upgrades Terraform.

Also worth a small guard: `setPrereqs` runs unconditionally in the `.then`, so navigating away from Settings before the check resolves sets state on an unmounted component. Harmless under React 18, but a `cancelled` flag in the cleanup is two lines.

Suggested fix: cache the report (module-level promise, or lift it into a small context/hook shared with the wizard) and reuse it across mounts.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
CoderCoco: Fixed in 9a3fbd2. Added `ElectronStoreService.bootstrap` (persisted via `WizardController.saveState`/`getState`, mirrored in `gsd-api.ts`). First-run now saves `{ bootstrap: resourceNames }` when leaving the bootstrap step; Reconfigure's prefill effect rehydrates it into `resourceNames` so `backendConfig` (and `terraform init`) uses the operator's actual names, not the defaults.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
CoderCoco: Fixed in 9a3fbd2. `commitReconfigureAnswers` now builds its `wizard.state.save` payload keyed off `completedSteps` — only a step actually opened via Edit (removed from that set) contributes its field. A step left collapsed (including one where the prefill silently failed) is omitted from the payload entirely, so `saveState` never touches its stored value. Added a regression test ("should not clobber stored config on Finish when the prefill itself fails and nothing was edited").

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
CoderCoco: Fixed in 9a3fbd2. The effect now waits until the `listAwsProfiles` fetch has settled (`profiles !== null || profilesError !== null`) before applying — avoids the transient "first pass sees `profiles === null`" misclassification — and is guarded by a `prefillAppliedRef` so it only ever applies once, regardless of what else changes `profiles` afterward.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
CoderCoco: Fixed in 9a3fbd2 — the button now only renders when `mode === 'reconfigure' && onCancel`.

app/packages/web/src/components/first-run-wizard/first-run-wizard.component.tsx
CoderCoco: Fixed in 9a3fbd2. Reworded the `mode` prop's TSDoc to scope the guarantee to the durable answer fields (`activeCloud`/`aws`/`bootstrap`) and call out explicitly that the paste-flow's `saveCredentials` call and the bootstrap step's Run bootstrap/Check permissions calls are real IPC side effects that happen immediately in both modes and aren't undone by Cancel.

app/packages/web/src/pages/settings.page.tsx
CoderCoco: Declined — leaving as-is. `PrerequisiteService.check()` is two local `execFile` calls (`terraform -version`, `aws --version`), not network I/O; `FirstRunWizard` itself already calls `wizard.prereqs.check()` unconditionally on every mount in `'first-run'` mode with no caching layer, so this matches existing behavior rather than introducing a new cost pattern. Settings is also not a hot-path route (unlike the dashboard's polling loop), so re-running two sub-100ms local processes per visit isn't worth adding a cache/staleness story for right now. Happy to revisit if it shows up as an actual UX or resource issue.

Issue #213, opened 2026-05-11
chore(terraform): migrate project_name default from "game-servers-poc" to "hyveon"


CoderCoco:
> Follow-up from the Hyveon rebrand (#212). Tracking the deferred half of the brand migration.

## Scope

Change `terraform/variables.tf`'s `project_name` default from `\"game-servers-poc\"` to `\"hyveon\"` (or another approved value), then plan and apply the resulting infrastructure churn.

## Why this is a separate PR

\`project_name\` is interpolated into resource names, IAM identifiers, log group names, tag values, and Discord secret ARNs. Changing it isn't a text replacement — it forces AWS to:

- Re-create resources whose name is not in-place editable (CloudWatch log groups, Secrets Manager secrets, possibly ECS task definitions).
- Re-tag every resource that carries \`Project = var.project_name\`.
- Break the **Cost Explorer 'Project' tag activation** that's already configured in AWS Billing — the activation is per-tag-value, so a new value needs a fresh activation and historical cost data won't carry across.

So this needs a dedicated planning pass, not an inline edit during a rebrand.

## Checklist

- [ ] Decide the new \`project_name\` value (\`hyveon\` is the obvious default; confirm).
- [ ] Read \`terraform plan\` output and enumerate every resource that will be replaced vs. retagged.
- [ ] Coordinate downtime if any in-flight game-server task definitions need to be redeployed.
- [ ] After apply, **re-activate the new \`Project = hyveon\` cost allocation tag** in AWS Billing → Cost allocation tags. Historical \`game-servers-poc\` cost data stays under the old activation.
- [ ] Update CLAUDE.md's reference to \`Project=game-servers-poc\` to the new value.
- [ ] Update \`terraform/terraform.tfvars.example\` if it sets the variable.
- [ ] Verify Discord Secrets Manager secrets came back online with the new ARN paths (the followup Lambda reads them at cold-start).
- [ ] Confirm the watchdog Lambda still finds tasks by tag.
- [ ] Sanity-check Cost Explorer breakdowns the next day.

## Out of scope

The \`@gsd/*\` → \`@hyveon/*\` workspace scope rename — tracked separately in #142.

PR #340, opened 2026-07-27
refactor: rename Gsd/gsd/GSD identifiers to Hyveon

CoderCoco:
Implements OpenSpec change `rename-gsd-to-hyveon` (proposal/design/spec/tasks under `openspec/changes/rename-gsd-to-hyveon/`), plus a follow-up branding sweep for the unabbreviated "Game Server Deploy(er)" name and PR-review fixes.

## Summary

Pure identifier rename, zero behavior change — `Gsd`/`gsd`/`GSD` → `Hyveon`/`hyveon`/`HYVEON` everywhere it means "this app," following the repo's existing `project_name` rebrand (#213) into the parts of the codebase Terraform doesn't touch:

- **`@hyveon/desktop-preload`**: `gsd-api.ts` → `hyveon-api.ts`, every `Gsd*` interface → `Hyveon*` (`GsdApi`, `GsdTestApi`, `GsdMockNamespaces`, and 12 per-domain `Gsd*Api` types), `contextBridge.exposeInMainWorld('gsd', ...)` → `exposeInMainWorld('hyveon', ...)`.
- **`@hyveon/web`**: every `window.gsd` call site and `Gsd*` type import → `window.hyveon` / `Hyveon*`, across `globals.d.ts`, `window.d.ts`, `api.service.ts`, and all consuming components/pages.
- **E2E test infrastructure**: `gsd-http-bridge.ts` → `hyveon-http-bridge.ts`, `applyGsdMocks()`/`installGsdHttpBridge()` → `applyHyveonMocks()`/`installHyveonHttpBridge()`, fixtures, page objects, and specs updated.
- **`@hyveon/desktop-main` + scripts + local dev tooling**: `.gsd/tfvars-bucket` marker → `.hyveon/tfvars-bucket` (with a legacy `.gsd` fallback + warning, added per review — see below), `GSD_TFVARS_BACKEND` env var → `HYVEON_TFVARS_BACKEND`, `.gitignore`, `electron-builder.yml` `appId: dev.gsd.desktop` → `dev.hyveon.desktop`.
- **Docs**: `CLAUDE.md`, `docs/docs/**`, and the live OpenSpec capability spec `openspec/specs/cloud-bootstrap/spec.md` updated. The dated historical design doc (`docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md`) is deliberately left untouched as a point-in-time record.

### Follow-up: unabbreviated "GameServerDeploy" branding

The repo hasn't shipped its first release yet, so unlike a typical rename, AWS-resource-named identifiers were renamed for real rather than just flagged:

- IAM policy `GameServerDeployAll` → `HyveonDeployAll` (`docs/docs/setup.md`, including its 3 statement `Sid`s and every prose cross-reference), plus the `GAME_SERVER_DEPLOY_ALL_ACTIONS` → `HYVEON_DEPLOY_ALL_ACTIONS` constant and every consumer (`iamPolicy.ts`, `IamCheckService.ts`, `wizard.controller.ts`, `hyveon-api.ts`, and the user-facing `<code>GameServerDeployAll</code>` string in the first-run wizard's bootstrap step)
- Root npm workspace renamed `game-server-manager` → `@hyveon/app` (9 `-w` targets in root `package.json` updated to match)
- `scripts/init-parent.ts`'s scaffolder prompt default drifted to the old `project_name = 'game-servers'` — corrected to `'hyveon'` (Terraform's own default was already `hyveon`)
- 6 test fixtures using `project_name = "game-servers"` moved to `"hyveon"`
- Left untouched: `docs/docs/guides/project-name-migration.md` (historical runbook naming the old default on purpose), the `game_servers` Terraform map / `{game}-server` ECS naming / `GameServer` TS domain type (these name the *hosted game*, not the product), and one `GameServerDeployAll` mention in the same excluded historical design doc above (kept consistent with that doc's existing exclusion).

### PR review fixes (4 threads, all resolved)

1. 19 stale `gsd` refs surfaced in `openspec/specs/**` after a `main` merge — synced to `hyveon`, added `MODIFIED Requirements` delta blocks.
2. Silent S3 tfvars-backend downgrade risk — `ConfigService.ts` / `scripts/tfvars-sync.ts` now fall back to the legacy `.gsd/tfvars-bucket` marker with a one-time warning log, instead of silently behaving as if no backend were bootstrapped.
3. Already-scaffolded parent-repo `Makefile` template hardcoded to `.gsd/` — now accepts either `.gsd` or `.hyveon` markers, with `TFVARS_MARKER_NEW` as the canonical write target going forward.
4. `appId` rename risk (breaks NSIS/macOS in-place upgrade for already-packaged installs) — documented as an accepted, pre-release risk rather than reverted.

## Test plan

- [x] Repo-wide case-insensitive grep for `gsd`/`GameServerDeploy`/`game-server-manager` (excluding `node_modules`, `.git`, build output, lockfiles, and the two excluded historical docs) returns zero residual hits
- [x] `npm run app:build` — green
- [x] `npm run app:lint` — green (0 errors, 3 pre-existing warnings confirmed present on `main`)
- [x] `npm run app:test` — green, 130/130 files, 2001/2001 tests
- [x] `npm run app:test:e2e` — green, 83/83 specs (chromium + electron projects)
- [ ] Manual smoke-test (`npm run app:dev`, confirm `window.hyveon` populated / `window.gsd` undefined in devtools) — not verifiable headlessly, left for a maintainer with a display


openspec/changes/rename-gsd-to-hyveon/tasks.md
```diff
@@ -1,49 +1,55 @@
...
 
 ## 6. Final verification
 
-- [ ] 6.1 Run a repo-wide case-insensitive `gsd` grep excluding `node_modules`, `.git`, build output dirs, `package-lock.json`, `docs/package-lock.json` (false-positive hash substrings), and `docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md` (intentionally-excluded historical doc) — confirm zero remaining hits.
-- [ ] 6.2 Run `npm run app:build`, `npm run app:lint`, `npm run app:test`, and `npm run app:test:e2e` (both projects) end to end; confirm all green.
-- [ ] 6.3 Manually smoke-test the packaged/dev Electron app once (`npm run app:dev`) and confirm `window.hyveon` is populated in the renderer devtools console, with `window.gsd` undefined.
+- [x] 6.1 Ran a repo-wide case-insensitive `gsd` grep excluding `node_modules`, `.git`, build output dirs, `package-lock.json`, `docs/package-lock.json`, `docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md`, **and `openspec/changes/**` (see scope note above)** — zero remaining hits. The only remaining hit outside those exclusions was inside `.git/` internal refs (the branch name `chore/rename-gsd-to-hyveon` itself), which is not repo content.
```
CoderCoco, line 53: **The "zero remaining hits" result is stale — it predates the `main` merge in this PR, and the live spec set now contradicts the shipped code.**

When this grep ran (at `8edd589`), `openspec/specs/` did not exist on the branch — `git ls-tree -d 8edd589 openspec/` returns only `openspec/changes`. The merge commit `08d7cdd` brought it in from `main`, and the `openspec/specs` tree on this branch is byte-identical to `main`'s (`4e8fdf3`), so nothing in it was renamed. `git grep -i gsd refs/remotes/origin/pr340 -- openspec/specs` returns 19 hits across 7 capability specs:

| Spec | Hits |
|---|---|
| `aws-credentials/spec.md` | 4 |
| `terraform-plan-apply-page/spec.md` | 4 |
| `terraform-run-history/spec.md` | 4 |
| `prerequisite-detection/spec.md` | 3 |
| `terraform-destroy-flow/spec.md` | 2 |
| `desktop-only-operator-surface/spec.md` | 1 |
| `wizard-flow/spec.md` | 1 |

This is not the same category as the excluded `openspec/changes/**` records. `openspec/specs/` is the *current* requirement set — "what the system does now" — so leaving it is not preserving history, it's letting the source of truth go stale. Two entries are now flatly wrong:

- `openspec/specs/aws-credentials/spec.md:30` — "the ad-hoc profile name defaults to `gsd-pasted`", and `:35` asserts `creds.aws.gsd-pasted`. This PR changes `AwsProfileService.DEFAULT_PASTED_PROFILE_NAME` to `hyveon-pasted`. The spec and the code now disagree on a named default.
- `openspec/specs/prerequisite-detection/spec.md:49` — mandates the surface be "mirrored in the preload as `gsd.wizard.checkPrereqs()` with a typed entry in `gsd-api.ts`". `gsd-api.ts` no longer exists.

Concrete failure: a maintainer runs `openspec` validation or drafts the next change against `openspec/specs/aws-credentials`, implements to the stated `gsd-pasted` default, and ships a wizard that writes to a store key the rest of the app doesn't read. Less dramatically, every future proposal that cites `gsd.terraform.runs.list` or `gsd-api.ts` is citing an API that isn't there.

Suggested fix: rename the 19 references in `openspec/specs/**`, and add `## MODIFIED Requirements` blocks to this change's delta spec (currently `specs/preload-bridge-naming/spec.md` only) covering the affected capabilities, so the sync leaves an audit trail rather than an unexplained in-place edit. Then re-run 6.1 at the merged head and update this line's result — the current text reads as a green check on a state that no longer exists.

app/packages/desktop-main/src/services/ConfigService.ts
```diff
@@ -537,7 +537,7 @@ export class ConfigService {
   private findTfvarsBucketMarker(startDir: string): string | undefined {
     let dir = startDir;
     while (true) {
-      const markerPath = join(dir, '.gsd', 'tfvars-bucket');
+      const markerPath = join(dir, '.hyveon', 'tfvars-bucket');
```
CoderCoco, line 547: **Existing operators silently lose their S3 tfvars backend — and the migration hint the design asked for was dropped, not relocated.**

`design.md` task 4.4 specified: *"if `.gsd/` exists and `.hyveon/` doesn't, print a manual-rename hint before continuing."* That hint was attached to `setup.sh`. `setup.sh` doesn't exist in this repo, so 4.4 was correctly marked N/A — but the *hint* went N/A with it, rather than moving to the surfaces that actually read the marker (this walk-up, and `findBucketMarker()` in `scripts/tfvars-sync.ts:281`). The requirement was hosted on a dead file and died with it.

The walk-up now matches only `.hyveon/tfvars-bucket`, returns `undefined`, and `getTfvarsBucket()` returns `null` with no log line — the `logger.warn` at :524 only fires on a *read* failure of a marker that was found.

Concrete failure: an operator who ran `init-parent bootstrap --s3-tfvars` before this change has `.gsd/tfvars-bucket`. They pull this change and open the app. `getTfvarsBucket()` → `null` → `GamesWriteService.ts:205` reports `mode: 'local'`, and `cloud-provider.module.ts:60`/`:93` construct the tfvars stores with `bucket: ''`. The app now writes `terraform.tfvars` locally and reports itself as having no S3 backend, while the bucket and its versioned objects still exist in S3. Nothing in the UI or the logs says "your marker moved" — the app just behaves as a fresh, un-bootstrapped install.

Suggested fix: have `findTfvarsBucketMarker()` fall back to `.gsd/tfvars-bucket` when `.hyveon/tfvars-bucket` is absent, and `logger.warn` once telling the operator to `mv .gsd .hyveon`. Mirror it in `findBucketMarker()` so the CLI and the app stay in agreement (their doc comments already promise they resolve identically). One release of legacy read is cheap; a silent backend downgrade is not.

scripts/init-parent.ts
```diff
@@ -266,14 +266,14 @@ STAMP_DIR   := $(REPO_ROOT)/.make
 # even before \`setup\` gets a chance to write its own submodule-local marker.
 # TFVARS_LOCK is the sidecar lock file tfvars-sync.ts writes after every
 # successful pull/push, recording the S3 version id/etag last synced.
-PARENT_TFVARS_MARKER := $(REPO_ROOT)/.gsd/tfvars-bucket
-TFVARS_MARKER := $(SUBMODULE)/.gsd/tfvars-bucket
+PARENT_TFVARS_MARKER := $(REPO_ROOT)/.hyveon/tfvars-bucket
```
CoderCoco: **Already-scaffolded parent repos keep a `Makefile` hardcoded to `.gsd/`, and the PR body's one-line migration doesn't fix it.**

This template renders `PARENT_TFVARS_MARKER`/`TFVARS_MARKER` and every `HYVEON_TFVARS_BACKEND`/`HYVEON_TFVARS_BUCKET` reference into a `Makefile` that lives in the operator's **private parent repo**, not here. Existing parent repos already have the previous render, with `.gsd/tfvars-bucket` and `GSD_TFVARS_BACKEND` baked in. Updating the submodule doesn't touch it.

That makes the PR body's stated migration (`mv .gsd .hyveon`) actively wrong for the primary supported topology. Concrete failure: operator reads the PR note, runs `mv .gsd .hyveon`. The desktop app and `tfvars-sync` are now happy. `make plan` is not — `$(wildcard $(PARENT_TFVARS_MARKER))` misses, `TFVARS_BACKEND` (`:290`) evaluates to `local`, and the pre-plan auto-pull at `:454` is skipped. Terraform then plans against whatever stale `terraform.tfvars` is on disk instead of the S3 copy, with no error — the recipe's whole design is that local mode is *silent*. If they instead skip the `mv`, the Makefile keeps working (it passes `--bucket` explicitly at `:302`) while the app reports no backend. Either way the two halves disagree, and neither half says so.

There's also no clean way out. `writeIfSafe` (`:672`) skips an existing `Makefile` without `--force`; `bootstrap --force` overwrites `terraform.tfvars` and `.gitignore` alongside it; `migrate --to-s3` writes the Makefile unconditionally but also flips the backend as a side effect. "Regenerate just the Makefile" isn't a supported operation.

Suggested fix: pick one and document it in `docs/docs/guides/s3-tfvars.md` (which currently documents only the post-rename world, as if it had always been that way) —
1. a `regen-makefile` subcommand that rewrites only the Makefile, plus a migration section stating both steps (`mv .gsd .hyveon` **and** regenerate), or
2. render the marker lookups to accept either directory for a release (`$(wildcard $(REPO_ROOT)/.hyveon/tfvars-bucket $(REPO_ROOT)/.gsd/tfvars-bucket)`), which makes an un-migrated Makefile keep working.

Either way this belongs in the repo's docs, not only in a PR description — a PR body is not somewhere an operator upgrading a submodule will ever look.

electron-builder.yml
```diff
@@ -1,4 +1,4 @@
-appId: dev.gsd.desktop
+appId: dev.hyveon.desktop
```
CoderCoco, line 1: **`appId` change breaks in-place upgrade for anyone already running a packaged build — and the risk pointer that flagged it was deleted rather than answered.**

The original task 4.6 read *"(see design.md's Risks section on `appId` impact for already-packaged installs)"*. The completed version drops that parenthetical with no resolution recorded, so the risk the design raised now reads as if it never existed.

What actually changes: electron-builder derives the NSIS uninstall GUID from `appId` when no explicit `guid` is set, and `appId` is the macOS `CFBundleIdentifier`. A new installer therefore registers as a *different product* — it installs alongside the old one rather than upgrading it, and the previous entry is orphaned in Add/Remove Programs. **I reasoned this from electron-builder's documented GUID derivation; I did not build an installer and observe it.**

Blast radius is genuinely small: `gh release list` is empty, so there are no published installs — this only bites locally-installed dev/test builds. What I checked and it is *not* a problem: `app.getPath('userData')` keys off `productName` (`Hyveon`, unchanged), so electron-store data stays put; safeStorage keys off product name / OS user rather than the bundle id, so encrypted pasted credentials survive.

Suggested fix: nothing in the code — just put the resolution back where the risk was raised. A line in `tasks.md` 4.6 saying "no releases exist; existing local installs must be uninstalled manually before installing a `dev.hyveon.desktop` build" closes it honestly. Alternatively pin `nsis.guid` to the old value if preserving the Windows upgrade path is worth more than the identifier consistency.

CoderCoco (commented):
## Headline

**The "zero `gsd` hits" claim is no longer true at the head SHA.** The merge commit `08d7cdd` pulled `openspec/specs/**` in from `main` — that directory did not exist on the branch when task 6.1's grep was run (`git ls-tree -d 8edd589 openspec/` returns only `openspec/changes`). The live spec set now carries **19 `gsd` references across 7 capability specs**, and the `openspec/specs` tree on this branch is byte-identical to `main`'s (`4e8fdf3`), so none of them were touched. Two of them now directly contradict the code this PR ships.

Everything else I found is migration-path shaped: the rename moves three pieces of *operator-persisted* state (`.gsd/` marker dir, the generated parent `Makefile`, the packaged `appId`) with no fallback, no in-repo migration note, and — in the marker's case — a migration hint that the design explicitly asked for and that quietly evaporated when task 4.4 was marked N/A.

## Findings

| # | Severity | Finding |
|---|----------|---------|
| 1 | **Fix before merge** | `openspec/specs/**` still specifies `gsd.*` IPC calls, `gsd-api.ts`, and `creds.aws.gsd-pasted` — the live spec now contradicts the shipped code |
| 2 | **Fix before merge** | `.gsd/tfvars-bucket` → `.hyveon/tfvars-bucket` has no legacy read and no hint; an existing operator silently loses their S3 tfvars backend |
| 3 | **Fix before merge** | Already-scaffolded parent repos keep a `Makefile` pointing at `.gsd/`/`GSD_TFVARS_BACKEND`; the PR body's `mv .gsd .hyveon` is not a sufficient migration, and no clean regen path exists |
| 4 | Cheap/optional | `appId` change breaks in-place upgrade of any locally-installed build; task 4.6's risk pointer was dropped rather than answered |

## What's solid

The preload rename itself is clean and complete: `hyveon-api.ts`, the `HyveonApi`/`HyveonTestApi`/`HyveonMockNamespaces` types, `exposeInMainWorld('hyveon', …)`, and the `HYVEON_TEST_MODE`-gated `__test` seam all line up, and the delta spec in `specs/preload-bridge-naming/spec.md` pins the `window.gsd is undefined` assertion rather than just the positive case. The read path for pasted credentials resolves by the *persisted* profile name (`aws.profile` + `creds.aws.<name>`), not by `DEFAULT_PASTED_PROFILE_NAME`, so flipping that constant to `hyveon-pasted` does **not** orphan existing stored credentials — I checked this specifically because it looked like a data-loss candidate and it isn't. Logging the ~30 out-of-scope files inline in `tasks.md` rather than silently folding them in is the right call, as is flagging the OpenSpec-changes exclusion as a judgment call instead of presenting it as pre-approved.

## What I verified vs. what I took as stated

**Verified myself:** the repo-wide `gsd` grep at the head SHA (fails — see finding 1); that `openspec/specs` arrived via the merge and is unmodified; that `package-lock.json`/`docs/package-lock.json` hits are base64 substrings inside `integrity` hashes and genuinely false positives; that no `window.gsd`, `Gsd*`, `.gsd/`, or `GSD_` reference survives outside `openspec/**` and the excluded superpowers doc; the pasted-credentials read path; `writeIfSafe`'s skip-unless-`--force` behaviour and the `migrate` path's unconditional Makefile write.

**Taken as stated, not re-run:** `npm run app:build`, `app:lint`, `app:test` (1995 tests), and `app:test:e2e` (83 specs). For a mechanical rename of this shape the compiler and the suite are the load-bearing check, and I did not re-execute them — I spot-checked the `@hyveon/web` and e2e hunks rather than reading all 71 files line by line.

**Not reviewed at all:** the Terraform side (untouched by this PR) and the docs prose beyond checking that the env-var/marker names in `s3-tfvars.md` and `submodule.md` match what the code and generated Makefile actually read (they do).

openspec/changes/rename-gsd-to-hyveon/tasks.md
CoderCoco: Confirmed — fixed in c74990e. Renamed all 19 stale `gsd` references across the 7 capability specs you listed (`aws-credentials`, `terraform-plan-apply-page`, `terraform-run-history`, `prerequisite-detection`, `terraform-destroy-flow`, `desktop-only-operator-surface`, `wizard-flow`), including the two flatly-wrong ones (`gsd-pasted` default, `gsd-api.ts` reference). Also added `## MODIFIED Requirements` delta blocks under `openspec/changes/rename-gsd-to-hyveon/specs/<capability>/spec.md` for each affected capability per your suggestion, so the sync has an audit trail. Updated `tasks.md` 6.1.1 recording the re-run grep (zero remaining hits in `openspec/specs/**`). `npm run app:build`/`app:lint`/`app:test` all green afterward (130/130 files, 2001/2001 tests).

app/packages/desktop-main/src/services/ConfigService.ts
CoderCoco: Confirmed — fixed in 066f934. `ConfigService.findTfvarsBucketMarker()` and `tfvars-sync.ts`'s `findBucketMarker()` now fall back to the legacy `.gsd/tfvars-bucket` path (with a warning naming the `mv .gsd .hyveon` fix) at each directory when `.hyveon/tfvars-bucket` isn't present, exactly as suggested. Neither write path changed — both only ever write the new `.hyveon` location, so this is read-only backward compat, not a permanent legacy write path. Added regression tests in `ConfigService.test.ts` and `tfvars-sync.test.ts` covering the fallback and the new-path-wins-when-both-exist case. `npm run app:build`/`app:lint`/`app:test` all green (130/130 files, 2001/2001 tests).

scripts/init-parent.ts
CoderCoco: Confirmed — fixed in d051cc0, went with your option 2 (accept either directory) since it keeps an already-scaffolded parent repo's Makefile working without a new subcommand or a second manual step, and it's the same read-fallback shape as the ConfigService/tfvars-sync.ts fix in the sibling thread. `PARENT_TFVARS_MARKER`/`TFVARS_MARKER` in the generated Makefile now resolve to the new `.hyveon` path if present, else the legacy `.gsd` path if that's present, else the new path. Added a `TFVARS_MARKER_NEW` variable as the canonical write target so `make setup` always converges a freshly-bootstrapped submodule marker onto the new `.hyveon` location rather than perpetually rewriting the legacy one. Documented the fallback in `renderMakefile`'s doc comment and updated `scripts/init-parent.test.ts`/`init-parent.cli.test.ts` to pin the new literal template output. `docs/docs/guides/s3-tfvars.md` didn't need updating since operators on an un-migrated Makefile now need no manual step at all. `npm run app:build`/`app:lint`/`app:test` all green (130/130 files, 2001/2001 tests).

electron-builder.yml
CoderCoco: Agreed — no code change needed, just put the resolution back where the risk was raised. Added to `tasks.md` 4.6 in d051cc0: this changes the NSIS/macOS bundle identity (a build with the new `appId` installs alongside an old one rather than upgrading it), `gh release list` is empty so no published installers are affected — only local dev/test installs, which need a manual uninstall before installing a `dev.hyveon.desktop` build — and `userData`/safeStorage both key off `productName` (unchanged), so stored data and encrypted credentials are unaffected. Not pinning `nsis.guid` to the old value, per your reasoning that identifier consistency is worth more than preserving an in-place upgrade path for unreleased dev builds.

CoderCoco:
## Fixes Applied Successfully

Fixed 7 file(s) based on 7 CodeRabbit feedback item(s) (the 8th, an "afterwards"→"afterward" spelling nit in `scripts/README.md`, was left for a separate decision — not applied here).

**Files modified:**
- `.gitignore`
- `app/packages/desktop-main/src/services/ConfigService.ts`
- `app/packages/desktop-main/src/services/ConfigService.test.ts`
- `scripts/tfvars-sync.ts`
- `docs/docs/guides/submodule.md`
- `docs/docs/guides/s3-tfvars.md`
- `scripts/README.md`

**Commit:** `9aa0c5f`

Verified: `npm run app:build`, `npm run app:lint` (0 errors), `npm run app:test` (130/130 files, 2001/2001 tests) all green after these changes.

scripts/README.md
```diff
@@ -46,28 +46,28 @@ a `--` flag) exits `1` with `Unknown subcommand "<token>"`.
...
   that already has a `Makefile`, i.e. `bootstrap` has already run) from a
   local-file tfvars backend onto a versioned S3 bucket. Writes
-  `.gsd/tfvars-bucket` (`${project_name}-tfvars`, read out of the existing
+  `.hyveon/tfvars-bucket` (`${project_name}-tfvars`, read out of the existing
   `terraform.tfvars`'s `project_name` key), rewrites the `Makefile` with the
-  S3-aware targets, then runs `make setup` with `GSD_TFVARS_BACKEND=s3` so
+  S3-aware targets, then runs `make setup` with `HYVEON_TFVARS_BACKEND=s3` so
   its self-contained S3 tfvars-bootstrap step (via `terraform/bootstrap/`)
   provisions the bucket. `terraform.tfvars` itself is left untouched — pull
   it back down from S3 afterwards with `make tfvars-pull` if you want
```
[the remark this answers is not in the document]
CoderCoco: Fixed in 2562d33.

Issue #275, opened 2026-07-12
chore(desktop): design and wire up a Hyveon app icon


CoderCoco:
> Companion housekeeping item for the [Hyveon desktop pivot initiative](https://github.com/CoderCoco/Hyveon/issues/214) — cleanup/polish piece alongside M4 (#140, #141).

## Problem

Hyveon has no app icon. `electron-builder.yml` sets no `icon` for any platform target (`win`, `mac`, `linux`), so packaged installers (`.exe` / `.dmg` / `.AppImage`) and the running app ship with electron-builder's generic default icon. There's also no favicon for the web/renderer (`app/packages/web/index.html` has no `<link rel="icon">`).

## Scope

- Design (or source) a Hyveon app icon and produce the platform-specific asset set electron-builder expects:
  - `build/icon.icns` (macOS)
  - `build/icon.ico` (Windows)
  - `build/icon.png` (Linux, ≥512×512) or a `build/icons/` directory of sizes
- Wire the assets into `electron-builder.yml` (`win.icon`, `mac.icon`, `linux.icon`, and/or the top-level `icon` field).
- Add a matching favicon to `app/packages/web/index.html` for the dev/browser-preview experience.
- Confirm the icon renders correctly in: the packaged installer, the OS taskbar/dock while running, and (Windows) Explorer's file icon for the installed `.exe`.

## Acceptance

- `npm run desktop:package` produces installers with the Hyveon icon on all three platforms (or documents which were verified if cross-platform builds aren't available locally — CI matrix in #146 covers all three).
- The running app shows the Hyveon icon in the taskbar/dock, not the electron-builder default.
- The web dev server / preview tab shows a matching favicon.

## Out of scope

- Final brand/logo design approval — assumes a source logo asset is supplied or a placeholder is acceptable to unblock the packaging wiring; swapping the actual artwork later is a non-breaking follow-up.

Issue #214, opened 2026-05-11
initiative: pivot Hyveon to a standalone Electron desktop app


CoderCoco:
> The single tracking issue for the Hyveon desktop pivot. Authoritative scope, locked decisions, and migration plan live in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Vision

Convert Hyveon from a Nest+React web app (operator-driven, Docker-shipped, hand-run `terraform apply`) into a **single-binary cross-platform Electron desktop application** that owns the full deploy lifecycle — config editing, plan/apply/destroy, log streaming — while keeping AWS as the v1 cloud and leaving abstraction seams for GCP/Azure later.

## Milestones

Each milestone is independently releasable.

### M1 — Runs as Electron
- [ ] #135 — epic: Electron shell + build pipeline
- [ ] #136 — epic: IPC migration of Nest controllers

After M1: desktop app launches; every current feature works inside the Electron window via IPC. Operator still hand-runs `terraform apply` out of band.

### M2 — Manages config
- [ ] #137 — epic: cloud provider abstraction + AWS impl
- [ ] #80 — epic: remote tfvars storage — `RemoteFileStore` interface + AWS S3 impl
- [ ] #81 — epic: `TfvarsService` reads from `RemoteFileStore` in desktop-main
- [ ] #82 — epic: add/edit/remove games via desktop UI

After M2: operator can add/edit/remove game-server entries from inside the app; tfvars round-trip to S3 with optimistic locking.

### M3 — Runs terraform
- [ ] #138 — epic: local terraform orchestration
- [ ] #139 — epic: first-run wizard + credentials UX

After M3: fully self-managing. Fresh install on a clean machine reaches "dashboard ready" through the wizard; plan/apply/destroy entirely from the UI.

### M4 — Polish & ship
- [ ] #140 — epic: test migration to Playwright Electron
- [ ] #141 — epic: distribution + auto-update scaffolding

After M4: shippable. NSIS / DMG / AppImage from CI; `electron-updater` wired (disabled until signing).

### Companion housekeeping
- [ ] #213 — chore(terraform): migrate `project_name` from `"game-servers-poc"` to `"hyveon"` — should land before #138 so the runs table and run-log S3 keys are created under the new name.
[...]

Issue #202, opened 2026-05-10
ci: tag-triggered release workflow publishing artifacts to GitHub Releases


CoderCoco:
Child of #141 (epic: distribution + auto-update scaffolding). Authoritative scope in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Scope

A `.github/workflows/release.yml` that triggers on tag push (`v*`). Builds NSIS, DMG, AppImage on the corresponding runners (Epic A.5 supplies the matrix logic). Uploads artifacts to a draft GitHub Release named after the tag; tagging is required to flip a release public.

## Acceptance

- Tag push produces a draft release with three artifacts attached.
- Artifacts download and install on each OS.

CoderCoco:
Verified against acceptance criteria via \`.github/workflows/package.yml\`:

- **Trigger**: \`on.push.tags: ['v*']\` (plus \`pull_request\` for build verification on every PR).
- **3-OS matrix**: \`package\` job matrix builds \`ubuntu-latest\` → AppImage, \`macos-latest\` → dmg, \`windows-latest\` → nsis exe, each uploaded as a workflow artifact.
- **Release job**: gated on \`startsWith(github.ref, 'refs/tags/v')\`, downloads all three matrix artifacts and attaches them via \`softprops/action-gh-release@v2\` with \`draft: true\`.

No \`v*\` tag has been pushed yet (checked \`gh run list --workflow=package.yml\`), so the release job hasn't executed for real — the next tag push will exercise it end-to-end. Workflow logic satisfies the issue's acceptance criteria as written; closing per the \`add-distribution-updater\` OpenSpec change (D7).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #141, opened 2026-05-10
epic: distribution + auto-update scaffolding


CoderCoco:
## Epic G: distribution + auto-update scaffolding

> Part of the **Electron desktop pivot**. Authoritative scope and rationale live in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Scope

CI matrix produces `.exe` (NSIS), `.dmg`, `.AppImage` artifacts and pushes them to GitHub Releases on tag. `electron-updater` wired but disabled by default — a feature flag in electron-store flips it on once signing arrives. Documented unsigned-MVP install steps per OS (right-click Open on macOS, SmartScreen "More info → Run anyway" on Windows). Auto-update is explicitly out of scope for v1 release.

## Children

- [ ] G.1 — #202 `ci: tag-triggered release workflow publishing artifacts to GitHub Releases`
- [ ] G.2 — #206 `feat(desktop-main): electron-updater wired with disabled feature flag`
- [ ] G.3 — #207 `docs: unsigned-MVP install instructions per OS`
- [ ] G.4 — #209 `docs: code-signing roadmap (Apple Developer, Azure Trusted Signing)`

## Acceptance
- Tag-triggered GitHub Action publishes three artifacts to a Release.
- `electron-updater` initializes but does not perform update checks unless the feature flag is true.
- Install docs in `docs/` cover the unsigned-MVP UX on each OS.
- Telemetry around update checks is gated by the same flag (no silent network calls on first launch).

PR #342, opened 2026-07-27
feat(desktop): scaffold electron-updater, install & signing docs

### LINKED ISSUE #206: feat(desktop-main): electron-updater wired with disabled feature flag
Child of #141 (epic: distribution + auto-update scaffolding). Authoritative scope in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Scope

Add `electron-updater` to the main process initialization. Behind an `enableAutoUpdate` boolean in electron-store, default `false`. When false, the updater is constructed but never calls `checkForUpdates`. Telemetry around update checks is gated by the same flag — no silent network calls.

## Acceptance

- First launch produces zero outbound traffic to update servers.
- Flipping the flag (manual test only) shows update prompts as expected once we have a signed build.

### LINKED ISSUE #207: docs: unsigned-MVP install instructions per OS
Child of #141 (epic: distribution + auto-update scaffolding). Authoritative scope in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Scope

Write `docs/docs/install.md` covering: macOS right-click → Open (gatekeeper), Windows SmartScreen "More info → Run anyway", Linux AppImage chmod +x + run. Screenshots if reasonable. Link from the project README.

## Acceptance

- Doc renders in the docusaurus site.
- README has a top-level "Install" link.

### LINKED ISSUE #209: docs: code-signing roadmap (Apple Developer, Azure Trusted Signing)
Child of #141 (epic: distribution + auto-update scaffolding). Authoritative scope in the [design spec](https://github.com/CoderCoco/Hyveon/blob/main/docs/superpowers/specs/2026-05-10-electron-desktop-pivot-design.md).

## Scope

Write `docs/docs/code-signing-roadmap.md` summarizing the path to signed releases: Apple Developer account ($99/yr) + electron-builder notarize config; Azure Trusted Signing (~$10/month) for Windows; flow for enabling auto-update once both are in place. Cost + ownership notes for future maintainers.

## Acceptance

- Doc renders in docusaurus.
- Captures the key decisions, ownership, and current blockers.

CoderCoco:
Closes #206
Closes #207
Closes #209

## Summary

Completes the remaining scope of the `add-distribution-updater` OpenSpec change (initiative #214, epic #141) as **one bundled PR** rather than one PR per issue — per explicit operator direction to push the initiative's distribution piece through in a single change.

- **#206 — electron-updater wired, disabled by default.** New `desktop-main/src/updater.ts` exports `initUpdater(store)`: a no-op outside a real Electron main process, and a no-op whenever the new `enableAutoUpdate` electron-store flag (added to `AppStoreSchema`) is falsy — so a first launch with default settings makes zero outbound calls to any update feed. `main.ts`'s `bootstrap()` now returns the Nest microservice app so `electron-entry.ts` can resolve `ElectronStoreService` from the same DI graph the IPC controllers use, rather than constructing a second instance. `electron-builder.yml` gains a `publish: github` block (owner/repo explicit, since there's no `repository` field in `package.json` to infer from) so packaged builds carry the `app-update.yml` feed metadata electron-updater needs. `desktop:package` is pinned to `--publish never` so CI packaging (and local runs) never auto-publish — release publishing stays solely owned by the tag-triggered `package.yml` workflow.
- **#207 — unsigned-MVP install docs.** `docs/docs/install.md`: where to download release artifacts, why the unsigned-build OS warnings appear, and exact per-OS steps (macOS right-click → Open for Gatekeeper, Windows SmartScreen "More info → Run anyway", Linux `chmod +x`). Linked from `docs/docs/setup.md`'s packaged-app section and a new top-level Install row in `README.md`'s docs table.
- **#209 — code-signing roadmap.** `docs/docs/code-signing-roadmap.md`: Apple Developer Program (~$99/yr) + notarization plan for macOS, Azure Trusted Signing (~$10/month) for Windows, the auto-update-enablement flow once signing lands, and why `mac.hardenedRuntime: false` in `electron-builder.yml` is intentional and must stay until real signing exists.
- **Housekeeping — #202.** Verified `.github/workflows/package.yml` already satisfies #202's acceptance criteria (tag-triggered `v*` push, 3-OS matrix, draft release via `softprops/action-gh-release@v2`) and closed it directly with an evidence comment — no code change needed, no PR required per the change's design (D7).
- **#275 (app icon)** was already shipped independently via PR #334 before this session — no action needed here.

## Notes / anomaly flagged

Epic **#141 was closed on 2026-07-23**, before #202/#206/#207/#209 — its own children — were done. The closing PR (#308) explicitly said in its body that it "does not close" anything beyond adding OpenSpec scaffolding, so this looks like an accidental/premature close. This PR is what actually finishes #141's real scope; flagging here since #141 can't be meaningfully "re-closed" by this PR. Initiative #214 will be closed as a follow-up once this PR merges (task group 6 of the OpenSpec change), not automatically by this PR.

## Test plan

- `npm run app:test` — 2011/2011 passing (includes new `updater.test.ts`, extended `ElectronStoreService.test.ts`, `main.test.ts`, `electron-entry.test.ts` coverage for the updated `bootstrap()`/`initUpdater` wiring).
- `npm run app:lint` — 0 errors (3 pre-existing warnings in `ConfigService.ts`, unrelated to this change).
- `docs && npm run build` (Docusaurus) — succeeds; confirmed `build/install/index.html` and `build/code-signing-roadmap/index.html` render with correct titles/headings.
- Manual code inspection: `initUpdater` never imports `electron-updater` or calls `checkForUpdates` unless both the Electron guard and `enableAutoUpdate` flag are true — covered by dedicated no-op tests for the outside-Electron, flag-false, and flag-absent cases.


app/packages/desktop-main/src/electron-entry.ts
```diff
@@ -65,11 +67,13 @@ function createWindow(): void {
 
 app.whenReady().then(() => {
   bootstrap()
-    .then(() => {
+    .then((nestApp) => {
       if (isTestMode()) {
         console.log('[desktop-main] HYVEON_TEST_MODE active — test seam enabled');
       }
 
+      void initUpdater(nestApp.get(ElectronStoreService));
```
CoderCoco: **`void` here converts a real failure into an unhandled rejection.**

`initUpdater` ends in `await autoUpdater.checkForUpdates()`. electron-updater's `checkForUpdates()` (`out/AppUpdater.js:264-273`) emits `error` **and then rethrows**:

```js
.catch((e) => {
  nullizePromise();
  this.emit("error", e, `Cannot check for updates: ...`);
  throw e;
});
```

So the `autoUpdater.on('error', ...)` listener in `updater.ts` logs the failure, and the same error *also* propagates out of `initUpdater` into a promise nobody handles. The `.catch()` on the `bootstrap()` chain doesn't cover it — `void` detaches it deliberately.

Concrete failure: operator flips `enableAutoUpdate` to `true` in the store, launches the packaged app on a machine with no network (or the GitHub feed 404s / rate-limits — see the `latest*.yml` comment on the roadmap doc, which makes a 404 the *expected* case today). `checkForUpdates()` rejects → unhandled rejection in the Electron main process. Under Node's default `--unhandled-rejections=throw` that becomes an uncaught exception, i.e. a "A JavaScript error occurred in the main process" dialog on every launch, instead of the one log line the error handler implies. Same path if `await import('electron-updater')` itself ever fails.

I reasoned this from electron-updater's source rather than observing it — the flag is off by default, so it can't happen today. That's exactly why it should be closed now rather than found by whoever flips the flag.

Suggested fix: give the detached promise a terminal handler, or make `initUpdater` never reject —

```ts
initUpdater(nestApp.get(ElectronStoreService)).catch((err: unknown) => {
  console.error('[desktop-main] updater init failed:', err);
});
```

…or wrap the `checkForUpdates()` call in `updater.ts` in a `try/catch` and document a never-throws contract there (which is the stronger option, since it keeps the guarantee next to the code that owns it).

docs/docs/install.md
```diff
@@ -0,0 +1,60 @@
...
+[code-signing roadmap](./code-signing-roadmap.md) for the plan to change
+that). Every OS treats an unsigned binary as unverified and shows a warning
+before it'll run. That's expected — it doesn't mean the build is broken or
+tampered with, just that nobody has paid for and configured a signing
+certificate yet. The steps below get you past each OS's warning.
+
+## macOS — right-click → Open
+
+1. Double-click the downloaded `.dmg` and drag **Hyveon** into `Applications`.
+2. Gatekeeper blocks a plain double-click of an unsigned app the first time —
```
CoderCoco: **These macOS steps dead-end on macOS 15 (Sequoia) and later.**

Sequoia removed the Control-click → Open override for Gatekeeper. On macOS 15+, a downloaded app with no valid signature shows a dialog whose only buttons are **Move to Trash** / **Done** — there is no "Open" button to click, whether the user double-clicks or right-clicks. The supported path is now: try to open once (so the block is recorded), then **System Settings → Privacy & Security → scroll to the blocked-app notice → "Open Anyway"**, then authenticate and confirm.

Concrete failure: a user on macOS 15/26 downloads the `.dmg`, follows steps 2–3 verbatim, gets a dialog with no "Open" button, and concludes the build is broken — which is precisely the outcome this page exists to prevent.

Suggested fix: make Privacy & Security → "Open Anyway" the primary instruction, and keep right-click → Open only as a labelled "macOS 14 and earlier" fallback.

Secondary, and **not verified** — worth someone actually launching a CI `.dmg` on Apple Silicon before release: an arm64 app with no signature at all (not even ad-hoc) is refused outright as "damaged", which no Gatekeeper override recovers from. electron-builder normally ad-hoc signs when it finds no identity, so this is probably fine, but this page's whole premise is that the unsigned build *does* run once the user gets past the warning, and nothing here records that anyone has confirmed that end-to-end on macOS.

docs/docs/code-signing-roadmap.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Enabling auto-update once signing lands
+
+`electron-updater` is already wired (see the `auto-update-scaffold`
+capability) but stays **disabled by default** via the `enableAutoUpdate`
+electron-store flag, specifically because an unsigned updater would fail
+Gatekeeper checks on macOS regardless of what the update feed says. Once
+signing is in place:
+
+1. Confirm `publish: github` in `electron-builder.yml` is still correct (it
```
CoderCoco: **"No change required there" is the one instruction in this doc that will send the next maintainer down a dead end.**

`publish: github` gets `app-update.yml` embedded in the packaged app, which is the client half. The feed half is missing: `.github/workflows/package.yml` uploads artifacts by glob — `release/*.AppImage`, `release/*.dmg`, `release/*.exe` — and the `release` job attaches `dist/**/*` from exactly those. electron-updater resolves a GitHub-provider feed by fetching `latest.yml` / `latest-mac.yml` / `latest-linux.yml` (plus `.blockmap` files for differential download) from the release assets. None of those match the globs, so a `v*` tag produces a draft release with three installers and no update metadata.

Concrete failure: signing lands, someone follows step 1 ("confirm, no change required"), flips `enableAutoUpdate` per step 2, ships `v1.0.1` — and every client's `checkForUpdates()` 404s on `latest.yml` and reports "Cannot check for updates" forever. Combined with the unhandled-rejection issue on `electron-entry.ts:75`, that 404 is a startup error dialog, not just a log line.

Suggested fix: rewrite step 1 as a real task — widen the matrix globs to include `release/latest*.yml` and `release/*.blockmap` (and confirm electron-builder actually emits `latest*.yml` under `--publish never`; if it doesn't, the tag job has to run electron-builder with `--publish onTagOrDraft` instead of hand-attaching artifacts). Right now the doc asserts something that hasn't been checked, which is worse than listing it as open work.

app/packages/desktop-main/src/updater.ts
```diff
@@ -0,0 +1,42 @@
...
+    logger.info('[updater] update available', { version: info.version });
+  });
+  autoUpdater.on('update-not-available', () => {
+    logger.info('[updater] no update available');
+  });
+  autoUpdater.on('update-downloaded', (info: { version: string }) => {
+    logger.info('[updater] update downloaded', { version: info.version });
+  });
+
+  await autoUpdater.checkForUpdates();
```
CoderCoco: **`autoDownload` and `autoInstallOnAppQuit` are left at their defaults, and both default to `true`.**

The TSDoc above describes this function as wiring "`checkForUpdates` + update-event logging", and the four listeners are all pure logging — which reads as observe-only. It isn't: with `autoDownload: true`, a successful `checkForUpdates()` immediately starts downloading the installer in the background, and `autoInstallOnAppQuit: true` then installs it the next time the user quits, with no prompt anywhere in this code path.

Concrete failure: whoever flips `enableAutoUpdate` to `true` — following the roadmap's step 2, which frames it as "pick up updates automatically" — gets a silent multi-hundred-MB download plus a silent install-on-quit, on a build with no UI to surface or cancel either. That's a defensible product decision, but it should be a stated one, not an inherited default.

Suggested fix: set both explicitly right after `autoUpdater.logger = logger` (`autoDownload = false` if the intent is scaffold-and-observe, `true` with a comment if it's genuinely wanted), and add a test asserting the chosen value so a future electron-updater default flip can't change behaviour silently.

app/packages/desktop-main/src/electron-entry.test.ts
```diff
@@ -108,7 +120,8 @@ describe('electron-entry', () => {
     mockLoadURL.mockResolvedValue(undefined);
     mockLoadFile.mockResolvedValue(undefined);
     mockQuit.mockImplementation(() => undefined);
-    bootstrapMock.mockResolvedValue(undefined);
+    bootstrapMock.mockResolvedValue(fakeNestApp);
+    initUpdaterMock.mockResolvedValue(undefined);
```
CoderCoco, line 124: **`initUpdaterMock` is set up but never asserted against.**

The mock is hoisted, registered via `vi.mock('./updater.js')`, and reset here, but no `expect(initUpdaterMock)` exists anywhere in the file — so the only genuinely new wiring in `electron-entry.ts` (resolve `ElectronStoreService` off the returned Nest app, hand it to `initUpdater`) has no coverage. `updater.test.ts` covers `initUpdater` in isolation; nothing covers the call site.

Concrete failure: someone refactors `bootstrap()` back to `Promise<void>`, or reorders the `.then` so `initUpdater` runs before `app.listen()` resolves, or drops the call while moving startup code — the suite stays green either way, and the regression only shows on a packaged build with the flag on, which nothing in CI exercises.

Suggested fix: two assertions in the existing "app ready" test — `expect(initUpdaterMock).toHaveBeenCalledOnce()` and `expect(fakeNestApp.get).toHaveBeenCalledWith(ElectronStoreService)`. A third case where `initUpdaterMock.mockRejectedValue(new Error('boom'))` and the window still opens would also pin the fix for the `void` issue on `electron-entry.ts:75`.

docs/docs/install.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Install
+sidebar_position: 4
```
CoderCoco: **`sidebar_position: 4` collides with the Guides category, and the roadmap page's `5` collides with Components.**

The autogenerated sidebar sorts top-level `docs/docs` entries — files by front-matter `sidebar_position`, directories by `_category_.json.position`. Current positions: `intro` 1, `architecture` 2, `setup` 3, **`guides/` 4**, **`components/` 5**. This page takes 4 and `code-signing-roadmap.md` takes 5, so both new pages tie with a category.

Concrete failure: Docusaurus doesn't error on ties — it breaks them by its own secondary ordering, so "Install" can render after "Guides" and "Code-signing roadmap" after "Components", the opposite of the intent stated in the PR body (an install page users hit before anything else). It'll also shuffle unpredictably as pages are added. The PR's "confirmed the page renders in the autogenerated sidebar" check passes either way, since it renders — just not necessarily where intended.

Suggested fix: give the two new pages 4 and 5 and bump `guides/_category_.json` to 6 and `components/_category_.json` to 7.

CoderCoco (commented):
## Review — `feat(desktop): scaffold electron-updater, install & signing docs`

Headline: **the flag-off path is genuinely airtight, but the flag-on path is not survivable yet** — `checkForUpdates()` rethrows after emitting `error` (electron-updater `out/AppUpdater.js:264-273`), and `electron-entry.ts` voids the promise, so the first offline/404 update check surfaces as an unhandled rejection in the main process rather than the logged-and-degraded behaviour the `on('error')` handler implies. Separately, the release pipeline never publishes `latest*.yml`, so even after signing lands and the flag flips, no client can ever resolve an update — and the roadmap doc tells the next maintainer the opposite.

### Findings, ranked

| # | Where | What | Rank |
|---|---|---|---|
| 1 | `electron-entry.ts:75` | `void initUpdater(...)` swallows a rejection that electron-updater deliberately rethrows | **Fix before merge** |
| 2 | `docs/docs/install.md:31` | macOS right-click → Open no longer bypasses Gatekeeper on macOS 15+; the documented steps dead-end | **Fix before merge** |
| 3 | `docs/docs/code-signing-roadmap.md:77` | Release job uploads only `*.exe`/`*.dmg`/`*.AppImage` — no `latest*.yml`, so the feed is unusable; the doc says "no change required there" | **Fix before merge** (doc + a workflow TODO) |
| 4 | `updater.ts:41` | `autoDownload` / `autoInstallOnAppQuit` left at their `true` defaults — flipping the flag silently downloads and installs on quit | Cheap |
| 5 | `electron-entry.test.ts:124` | `initUpdaterMock` is wired but never asserted — nothing pins the entry → `initUpdater` wiring | Cheap |
| 6 | `docs/docs/install.md:3` | `sidebar_position: 4` / `5` collide with the Guides (4) and Components (5) `_category_.json` positions | Cheap |

### What's solid

- The double gate in `updater.ts` (`process.versions.electron` **then** the store flag **then** the dynamic `import`) is the right shape, and the three no-op tests cover exactly the cases that matter — outside Electron, flag false, flag absent. "Zero outbound calls on a default first launch" holds as claimed.
- `bootstrap()` returning the `INestMicroservice` instead of `electron-entry.ts` constructing a second `ElectronStoreService` is the correct fix; `ElectronStoreService` is a provider of `ElectronStoreModule`, which `AppModule` imports, so the non-strict `get()` resolves.
- `--publish never` pinned in `desktop:package` with the explaining comment in `electron-builder.yml` is the right division of labour against the tag-triggered release job.
- `enableAutoUpdate?: boolean` as optional-absent-means-off (rather than a defaulted `true`-able key) is the conservative choice and the round-trip tests cover both backings.

### What I verified, and what I didn't

- **Verified by reading**: `electron-updater@6.8.9`'s `checkForUpdates()` rethrow path; `AppModule`/`ElectronStoreModule` provider wiring; `externalizeDepsPlugin()` resolves against the root `package.json`, which declares no `dependencies`, so `electron-updater` is bundled into `out/main` rather than needing a `files:` entry like `@cdktf/hcl2json`; the `package.yml` artifact globs; the Docusaurus `_category_.json` positions; `clearMocks`/`restoreMocks` are on, so `updater.test.ts`'s `toHaveBeenCalledOnce` isn't order-dependent.
- **Taken as stated, not re-run**: the 2011-test result, the lint result, and the Docusaurus build. CI is green on this head — `test`, `build`, `e2e`, `integration`, `eslint`, and `package` on all three OS runners — and the mac/windows package logs show blockmaps being built, so packaging genuinely works with the new dependency.
- **Reasoned, not observed**: findings 1 and 3 cannot fire today because the flag is off by default, so neither is reproducible from a local run. Finding 1 rests on electron-updater's source and Node's default unhandled-rejection handling; finding 3 rests on the workflow globs. I also did **not** confirm that `app-update.yml` actually lands in the packaged resources under `--publish never` — I'm assuming it does, which is the behaviour the `publish:` block is there for.
- **Not reviewed**: nothing in the Terraform or Lambda tree is touched by this PR. I did not download a CI installer to sanity-check the packaged app.

One process note, not a code finding: `tasks.md` §5 marks 5.1–5.8 `[x]` including "Run `npm run desktop:package` locally and verify the installer …", crediting work done in PR #334. The prose note above the list says so, but the checkboxes on their own read as verification this PR performed.

app/packages/desktop-main/src/electron-entry.ts
CoderCoco: Fixed in 920a9ba, hardened further in 50f785d.

Went with the "stronger option" you named: `updater.ts` now wraps `await autoUpdater.checkForUpdates()` in try/catch, so `initUpdater` can no longer reject at all (the `'error'` listener already logged it). Added `updater.test.ts` coverage proving `initUpdater` resolves even when `checkForUpdates` rejects.

Also added the belt-and-suspenders `.catch()` at the `electron-entry.ts` call site (your other suggested option) as a second independent guarantee, plus a test (`electron-entry.test.ts`) proving the window still opens even if `initUpdater` rejects anyway.

docs/docs/install.md
CoderCoco: Fixed in 920a9ba. Confirmed via web search — macOS Sequoia did remove the right-click → Open bypass. Rewrote the macOS section: primary path is now System Settings → Privacy & Security → "Open Anyway" (first launch only, admin auth required), with a note that subsequent launches work normally.

Left the arm64/ad-hoc-signing verification point as-is for now — agreed it's unverified, but confirming a real CI `.dmg` boots on Apple Silicon needs actual macOS hardware, which isn't available in this session. Flagging it here rather than claiming something I haven't checked; happy to note it as an open item in the doc if you'd rather have it written down than just live in this thread.

docs/docs/code-signing-roadmap.md
CoderCoco: Fixed in 920a9ba. Confirmed locally: ran `npm run desktop:package` on Linux and `latest-linux.yml` is generated into `release/` even with `--publish never` — so no `onTagOrDraft` change needed, the gap was purely in what CI chose to upload.

Went further than the glob widening, too: while verifying I checked `MacUpdater.js` and found it explicitly ignores dmg/pkg and only ever resolves a `zip` asset (`findFile(files, "zip", ["pkg", "dmg"])`) — so macOS auto-update was unfixable via CI globs alone. Added `zip` as a second mac target in `electron-builder.yml`. `package.yml`'s matrix now uploads `latest*.yml` + `*.blockmap` per platform alongside each installer, and the roadmap doc's step 1 now states what's actually required instead of "no change required there."

app/packages/desktop-main/src/updater.ts
CoderCoco: Fixed in 920a9ba. Set both `autoDownload` and `autoInstallOnAppQuit` to `false` explicitly right after `autoUpdater.logger = logger`, with a comment explaining why (no consent UI exists yet). Added a test asserting both are `false` after `initUpdater` runs.

app/packages/desktop-main/src/electron-entry.test.ts
CoderCoco: Fixed in 920a9ba (+ the third case in 50f785d). Added `expect(fakeNestApp.get).toHaveBeenCalledWith(ElectronStoreService)` and `expect(initUpdaterMock).toHaveBeenCalledWith(fakeStore)` in a new test, plus the rejection-resilience case you suggested (`initUpdaterMock.mockRejectedValueOnce(...)`, asserting the window still opens) once the call site had a real `.catch()` to make that test meaningful rather than trigger a genuine unhandled rejection in the test run.

docs/docs/install.md
CoderCoco: Fixed in 920a9ba, though via the less invasive of the two options: moved the two new pages to positions 6/7 instead of bumping `guides/_category_.json` and `components/_category_.json` up to 6/7. Same collision-free result, without touching the two existing category files. Rebuilt the Docusaurus site locally to confirm the new positions render without ties.

CoderCoco:
@coderabbitai review

---

PR #344, opened 2026-07-27
feat(web): allow setting the HTTPS flag from the game form

CoderCoco:
Closes #335

## Summary

- `gameServerValidator`: mirror the four Terraform `https = true` port rules (at least one port; first port `tcp` exact/lowercase; every port `tcp`/`udp`; no port 80/443), gated on `https === true`, so a config the app accepts also passes `terraform validate`.
- `WizardDraft` gains an `https: boolean` field, threaded through `createEmptyWizardDraft`/`draftFromGameServer`/`draftToPayload`/`toProposedEntry` — shared by both the add-game wizard and the edit form.
- `NetworkingStep` renders the HTTPS checkbox (with an associated `Label`) and an amber warning callout (stack-wide 443/80 exposure, loss of this game's direct port ingress, first-boot ACME/DNS requirement) shown only while enabled; the no-ports rule (no port row to anchor to) surfaces beside the toggle instead.
- `EditGameForm` drops the old `https` carry-forward workaround now that the draft owns the value (`environment`'s carry-forward is untouched, since it's still not editable).
- `ReviewStep` shows the HTTPS flag in the wizard's Networking summary.
- `TfvarsService.write.test.ts` gains round-trip coverage for the omitted/false/true HCL emission cases, which previously had none.

## Test plan

- [x] `npm run app:test` — 2130/2130 passing
- [x] `npm run app:lint` — 0 errors (3 pre-existing warnings in `ConfigService.ts`, unrelated)
- [x] Manual pass against the packaged Electron app (Playwright `_electron` driver under `xvfb-run`, mocked `games.list`/`games.update` IPC):
  - Toggle renders unchecked by default, checked for an `https: true` game
  - Enabling it shows the amber callout with all three consequences
  - A `udp` first port blocks Save with the violation shown on that port row
  - Correcting the protocol to `tcp` clears the violation and re-enables Save
  - Saving returns to the read-only detail view with **HTTPS: Enabled**, with no manual refetch


CoderCoco (commented):
## Headline

The four HTTPS port rules are transcribed from `terraform/aws/variables.tf` faithfully (I diffed them line by line against the validation block, including the exact-lowercase `tcp` comparison and the 80/443 rejection on *every* port, not just `ports[0]`), and the shared-validator placement means the IPC/HTTP write surfaces are covered too. The edit-game flow works exactly as the spec describes.

**The add-game wizard does not.** `checkHttpsPortRules` is called inside the `if (parsed.success)` branch, and at the Networking step the draft's `volumes` array is still empty (Storage is the *next* step), so `gameServerSchema` fails structurally and every business rule — including all four HTTPS rules — is skipped. I confirmed this against the branch rather than reasoning about it:

```
validateStep('networking', { https: true, ports: [{container: 8080, protocol: 'udp'}], volumes: [] })
  -> []            // no violation, "Next" is enabled
validateStep('networking', { https: true, ports: [], volumes: [] })
  -> []            // the no-ports rule doesn't fire either
```

So tasks 3.3/3.4 and the spec requirement "the Networking step shows the violation against the offending port row" hold only in the edit form. In the wizard the violation first appears at Review — where `ReviewStep` isn't given `issues`, so the operator gets a disabled **Submit** with nothing on screen explaining why.

The *primary* guarantee still holds — a Terraform-invalid config can't be written, because `advanceDisabled` covers all issues at Review and `GamesWriteService` re-validates server-side. This is a validation-timing/UX defect, not a correctness hole.

## Findings

| # | Finding | Rank |
|---|---------|------|
| 1 | HTTPS rules skipped at the wizard's Networking step (empty `volumes` short-circuits every business rule) | Fix before merge |
| 2 | Review step shows a disabled Submit with no visible reason once a rule does fire | Fix with #1, or note as pre-existing |
| 3 | `role="alert"` on a persistent, non-dynamic callout | Cheap |
| 4 | Stray double blank line in `wizard-form.utils.ts` | Cheap |

## Credit where due

- Mirroring the rules in `validateGameServer` rather than the component (D3) is the right call, and the `terraform/aws/variables.tf` source-of-truth comment is exactly the mitigation the drift risk needs.
- D6's `=== true` normalisation with the `?? false` rationale spelled out in the code, not just the design doc, is the kind of note that stops someone "simplifying" it back in six months.
- Deleting the `https` carry-forward *and* the comments that described it — rather than leaving stale prose behind — is the part these refactors usually miss.

## What I verified vs. didn't

**Verified (ran it):**
- All 10 touched test files against the PR head in a detached worktree: **184 passed**.
- `openspec validate add-https-toggle-to-game-form --strict` → valid.
- The validator-short-circuit behaviour above, via a throwaway spec against the branch's real `validateStep`.
- Rule transcription against `terraform/aws/variables.tf:52-80`, and the callout's three claims against `terraform/aws/main.tf:64-112` (the `direct_game_ports` `if !cfg.https` filter and the `length(local.https_games) > 0` 443/80 block). All three are accurate.

**Not verified (taking as stated):**
- The full `npm run app:test` 2130/2130 and `npm run app:lint` runs — I only ran the touched files.
- The manual Electron pass in the test plan.
- I did not read the e2e specs, and did not check whether any Playwright spec asserts on the Networking step's DOM (the new toggle sits above the port rows and shifts the layout).

One accuracy nit on the callout wording, not worth a thread: "this game's raw container port loses its public ingress rule" is true per *port key*, not per game — `direct_game_ports` dedupes on `${port}-${protocol}`, so if a non-HTTPS game declares the same port, the 0.0.0.0/0 rule survives. Fine as operator-facing prose.

app/packages/shared/src/gameServerValidator.ts
```diff
@@ -341,6 +393,7 @@ export function validateGameServer(
     issues.push(...checkFargateCpuMemoryPairing(parsed.data));
     issues.push(...checkAbsolutePaths(parsed.data));
     issues.push(...checkConnectMessagePlaceholders(parsed.data.connect_message));
+    issues.push(...checkHttpsPortRules(parsed.data));
```
CoderCoco: **The HTTPS rules never run at the point the wizard needs them.**

This call sits inside `if (parsed.success)`, alongside the other cross-cutting rules. `gameServerSchema` requires `volumes` to have at least one entry (`.min(1)`), and in the add-game wizard `volumes` is filled in on the **Storage** step, which comes *after* Networking. So while the operator is on the Networking step, the parse always fails on `volumes`, and every rule in this branch — including all four HTTPS rules — is skipped.

Concrete failure (confirmed against this branch, not inferred):

1. Open Add game, fill Identity + Resources, reach Networking.
2. Tick **Enable HTTPS (Caddy sidecar)**, add one port `8080/udp`.
3. `validateStep('networking', draft, [])` returns `[]` — no red row, no message, **Next** stays enabled.
4. Same with zero ports: the `ports`-pathed "must declare at least one port" rule doesn't fire either, so the new error slot next to the toggle can't be reached from the wizard at all.
5. The violation only materialises at Review, after Storage supplies a volume.

`checkPortCollisions` already solves this — it runs *outside* the `parsed.success` gate, against the raw proposed entry, precisely so port feedback survives a structurally incomplete draft. The HTTPS rules have the same property: they only need `ports` (plus the `https` flag), never `volumes`/`cpu`/`memory`.

Suggested fix: hoist `checkHttpsPortRules` out of this branch and run it on the same structurally-validated `ports` slice `checkPortCollisions` uses, gated on `proposed.https === true`. Then add a `validateStep('networking', ...)` test with `volumes: []` — the current `networking-step.component.test.tsx` cases all inject `issues` by hand, so none of them would have caught this.

app/packages/web/src/components/add-game-wizard/add-game-wizard.component.tsx
```diff
@@ -230,7 +230,13 @@ export function AddGameWizard() {
...
+          <NetworkingStep
+            ports={draft.ports}
+            issues={stepIssues}
+            onChange={(ports) => patchDraft({ ports })}
+            https={draft.https}
+            onHttpsChange={(https) => patchDraft({ https })}
+          />
         )}
         {step === 'storage' && <StorageStep draft={draft} issues={stepIssues} onChange={patchDraft} />}
         {step === 'review' && <ReviewStep draft={draft} submitError={submitError} />}
```
CoderCoco: **Dead-end at Review: Submit is disabled with nothing on screen saying why.**

`advanceDisabled = stepIssues.length > 0`, and on the review step `stepIssues` is *every* issue in the draft — but `ReviewStep` only receives `draft` and `submitError`, so nothing renders the issue list.

This is pre-existing (the Fargate cpu/memory rule can already land here for the same reason), but the HTTPS rules make it reachable on a normal happy path: because of the short-circuit flagged on `gameServerValidator.ts`, an HTTPS + `udp` first port sails through Networking and Storage, and the *first* time the operator learns anything is wrong is a greyed-out **Submit** on the last step. The message exists and is correctly pathed at `ports[0]` — it's just only visible if they happen to click Back to Networking.

Fixing the short-circuit makes this much rarer (the rule then blocks at Networking, where it renders), which is why I'd fix that one first. If you want the belt-and-braces version, passing `issues={stepIssues}` into `ReviewStep` and rendering them with their paths would close the last gap — the design doc's own Open Question #2 raises this.

app/packages/web/src/components/add-game-wizard/networking-step.component.tsx
```diff
@@ -76,6 +94,45 @@ export function NetworkingStep({ ports, issues, onChange }: NetworkingStepProps)
...
+        {noPortsIssue && (
+          <p id="https-error" role="alert" className="text-xs text-[var(--color-red)]">
+            {noPortsIssue.message}
+          </p>
+        )}
+
+        {https && (
+          <div
+            id="https-callout"
+            role="alert"
```
CoderCoco: **`role="alert"` is the wrong role for this callout, and it double-announces.**

`role="alert"` is an assertive live region: it interrupts the screen reader whenever the node appears. Two consequences:

- Opening the edit form for a game already declared `https = true` mounts this div immediately, so the operator gets the full three-sentence warning read at them assertively on page load, for a state they didn't change.
- The same node is the `aria-describedby` target on the checkbox above, so it's announced again on focus — once as an interrupting alert, once as the control's description.

The `<p id="https-error">` above it is a genuine alert (it appears in response to a validation failure) and should keep the role. This one is persistent explanatory text.

Suggested fix: drop `role="alert"` here, or use `role="note"`. The `aria-describedby` wiring already does the work of associating it with the control, and the amber styling plus `AlertTriangle` carries the visual weight.

app/packages/web/src/components/add-game-wizard/wizard-form.utils.ts
```diff
@@ -221,9 +233,11 @@ function toProposedEntry(draft: WizardDraft): Record<string, unknown> {
             mode: seed.mode.length > 0 ? seed.mode : undefined,
           }))
         : undefined,
+    https: draft.https,
   };
 }
 
+
```
CoderCoco: Stray second blank line between `toProposedEntry` and the `checkImage` doc comment — an artifact of the `https: draft.https,` insertion. Every other declaration boundary in this file is single-spaced. One-character cleanup.

app/packages/shared/src/gameServerValidator.ts
CoderCoco: Fixed in 805f526 — hoisted `checkHttpsPortRules` out of the `parsed.success` branch, gated on raw `proposed.https === true`, running against the same structurally-validated `ports` slice `checkPortCollisions` uses. Added a regression test with `volumes: []` reproducing exactly the failure you confirmed, plus the no-ports case. Thanks for tracing this all the way to a concrete repro instead of just flagging the pattern.

app/packages/web/src/components/add-game-wizard/add-game-wizard.component.tsx
CoderCoco: Fixed in 805f526 — did the belt-and-braces version too: `ReviewStep` now accepts an `issues` prop and renders every outstanding validation issue as an alert above `submitError`, so a disabled Submit always has a visible reason. Wired `issues={stepIssues}` in at the review-step call site. Added a test covering the rendered list.

app/packages/web/src/components/add-game-wizard/networking-step.component.tsx
CoderCoco: Fixed in 805f526 — dropped `role="alert"` from the persistent callout. The per-rule `<p id="https-error" role="alert">` (a genuine validation alert) is unchanged. `aria-describedby` still associates the callout with the toggle.

app/packages/web/src/components/add-game-wizard/wizard-form.utils.ts
CoderCoco: Fixed in 805f526 — removed the stray blank line.

PR #354, opened 2026-07-29
docs: slim CLAUDE.md and add a docs-writing skill with reviewers

CoderCoco:
## Why

`CLAUDE.md` had grown to 283 lines, most of it architecture prose duplicating `docs/docs/architecture.md` and `docs/docs/components/*`, plus behaviour that is now captured as requirements under `openspec/specs/`. Duplicated content drifts, and several sections had already gone stale.

This rewrites it as **workflow rules and invariants only** (181 lines), leading with a routing table that points at `openspec/specs/`, `openspec/changes/`, and the docs site so an agent reads the source of truth instead of a summary of it.

## What changed

**`CLAUDE.md` — 283 → 181 lines**

Removed: the six architecture subsections, the Discord/slash-command walkthrough, and the inline Electron test-seam documentation. Each is now reachable from the routing table.

Kept and tightened: commands, a ten-bullet "invariants that hurt to break" list, code/test conventions, git + PR + review workflow.

Stale content corrected while rewriting:

- four Lambdas → five (`lambda-efs-seeder` was missing)
- `@hyveon/cloud-aws` and `@hyveon/desktop-preload` missing from the workspace list
- `app:typecheck`, `app:test:coverage`, `scripts:tfvars-sync` undocumented
- the Playwright conventions still told agents to import the `authGate` / `authedPage` fixtures and seed a bearer token into `localStorage` — both were removed along with the HTTP surface
- "Two-tier browser testing strategy" heading sitting over a three-tier table
- page-object list missing `GamesPage`, `AuditPage`, `TerraformPage`, `TerraformHistoryPage`
- routing table gained the `docs/docs/app/` operator walkthroughs and `docs/docs/install.md` from #345

**New: "Before opening a PR"** — the two rules the file never stated. Run `app:lint`, `app:typecheck`, `app:test`, plus `app:test:integration` / `app:test:e2e` when the relevant layers changed, *locally*, before opening a PR. Then confirm `docs/docs/**` is updated in the same PR, OpenSpec specs are synced or archived, and the five-file Terraform variable checklist is satisfied.

**Moved, not deleted:** the `window.hyveon.__test.mock` seam was the only content in `CLAUDE.md` with no home in `docs/`. It now lives in `docs/docs/components/integration-tests.md`, clearly labelled as tier-1 material (that page is otherwise tier-2), with a pointer from `CLAUDE.md`.

**New: `.claude/skills/write-docs/`** — orchestrates documentation work: map the diff to the pages that own it, draft through a writer subagent, then review through three read-only evaluators in parallel before applying findings.

| Agent | Lens | Model |
|---|---|---|
| `docs-writer` | Drafts one page; must verify every claim against code, tests, or specs in the same invocation; returns sources, not prose | sonnet |
| `docs-accuracy-auditor` | Is every claim true of the code as it exists now? | opus |
| `docs-coverage-auditor` | Did the change update everything it owed — owner page, components index, `CLAUDE.md` routing table, OpenSpec sync, the five-file Terraform checklist? | opus |
| `docs-style-reviewer` | Will it build and navigate correctly? Separates build-breaking from advisory | sonnet |

Repo specifics are baked into the agents rather than left to inference: `onBrokenLinks`/`onBrokenMarkdownLinks` are both `throw`; internal links are root-relative with no `/docs` prefix (`routeBasePath: '/'`); the sidebar is autogenerated so `sidebar_position` + `_category_.json` control ordering; `.md` parses as MDX so `{game}` / `Partial<T>` must be backticked; `docs/static/img/app/**` is generated by `npm run docs:screenshots`.

## Test plan

- No code changes — documentation and agent/skill definitions only, so the unit, e2e, and integration suites are unaffected by this diff.
- Every claim in the rewritten command block was checked against `package.json` / `app/package.json` rather than carried over.
- Workspace and Lambda counts checked against `app/packages/` and `app/packages/lambda/`.
- Removed fixture references verified absent: `grep -rn "authGate\|authedPage" app/packages/web/e2e/` returns nothing.
- Docusaurus behaviour (`onBrokenLinks`, `routeBasePath`, autogenerated sidebar) read from `docs/docusaurus.config.ts` and `docs/sidebars.ts`.
- **Not verified:** `npm --prefix docs run build` could not be used to validate the appended MDX — it fails SSG for all 11 pages on clean `main` too (Node v24.18.0 against Docusaurus 3.6.3, pre-existing and unrelated). The appended section uses only constructs already present in that file. Both the skill and `docs-style-reviewer` record this signature so a future agent doesn't mistake it for its own breakage — or report a build as passing when it never ran.

No linked issue.


.claude/agents/docs-accuracy-auditor.md
```diff
@@ -0,0 +1,65 @@
...
+by design — it routes to `docs/`. Every future agent trusts these pages. A
+plausible-sounding wrong sentence is the exact failure this audit exists to catch,
+and plausibility is precisely what makes it hard to spot on a read-through.
+
+## How to operate
+
+1. Get the changed docs content:
+   `git diff <base>...HEAD -- docs/docs/` (default base `origin/main`; fall back to
+   `git diff HEAD~1 -- docs/docs/` when there is no upstream). If the diff is
+   empty, say so and stop — do not go audit the whole site uninvited.
```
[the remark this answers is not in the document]
CoderCoco: Fixed in f3a10f6.

Partly accurate as filed: both auditors already took a `<base>` placeholder with `origin/main` as a *default*, not a hard-coded ref. The real gap was the handoff — the skill said "give each the diff range" without specifying a form, so an evaluator on a stacked branch could quietly fall back to the wrong base.

Applied:
- `SKILL.md` step 3 now passes `base ref: <ref>` explicitly, along with the touched pages and changed code files.
- Both auditors now use the ref they were dispatched with, fall back to `origin/main` only when none was given, and must state which base they used — an audit of the wrong range is indistinguishable from a clean one, which is exactly the failure worth guarding against.

.claude/agents/docs-accuracy-auditor.md
```diff
@@ -0,0 +1,65 @@
...
+
+```
+VERDICT: <N claims checked, M wrong, K unverifiable>
+
+WRONG
+- <claim, quoted> — <what is actually true> (evidence: path:line)
+
+UNVERIFIABLE
+- <claim> — <what you looked at and why it was inconclusive>
+```
```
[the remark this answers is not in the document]
CoderCoco: Fixed in f3a10f6 — added `text` to all three evaluator report fences.

Note this repo has no markdownlint config or CI job, so MD040 is not enforced here. Taking it anyway for consistency: `docs-style-reviewer` itself tells other agents to flag fences with a missing language tag, so it should not ship with three of its own.

.claude/skills/write-docs/SKILL.md
```diff
@@ -0,0 +1,148 @@
...
+Re-run `docs-accuracy-auditor` only if you changed factual claims while applying
+fixes. Re-running all three on a cosmetic edit is waste.
+
+### 5. Verify
+
+```bash
+npm --prefix docs run build     # catches broken links and MDX parse errors
+```
+
+Known breakage: on Node 24 with Docusaurus 3.6.3 this fails SSG for every page
```
CoderCoco: **This known-breakage note is false, and it disables the skill's only verification step.**

I ran it at this PR's head with a clean install:

```
cd <worktree at f3a10f6>/docs && npm ci && npm run build
[SUCCESS] Generated static files in "build".
```

Node `v24.18.0`, `@docusaurus/core@3.6.3` — exactly the combination named here. The site builds, including the section this PR appends to `integration-tests.md`.

The failure you saw is real but has a different cause: `docs/node_modules` in the main checkout sits at 3.6.3 while `docs/package.json` pins 3.10.2 after #352, and building against that mismatched tree fails SSG for every page with `Cannot read properties of undefined (reading 'id')` at `DocItem`. I reproduced that too. It is an incremental-install artifact. #352's test plan says the same thing outright: "3.6.3 was not broken — it builds fine under Node 24", verified on a clean `npm ci`.

Concrete failure: an agent follows this skill, writes `[Terraform](/docs/components/terraform)` with the `/docs` prefix the gotchas warn about, runs the build, sees it throw, reads this paragraph, confirms a *stale* `main` tree fails the same way, and reports "pre-existing, unrelated to my change" — shipping a broken link into a site configured `onBrokenLinks: 'throw'`. The note is also version-stale twice over: `main` is on 3.10.2 now, so the "upgrade in flight" is landed.

Suggested fix: replace the paragraph with the install precondition. Something like — run `npm ci` in `docs/` before building; a stale incremental `node_modules` (Docusaurus version out of step with `package.json`) fails SSG for every page with `Cannot read properties of undefined (reading 'id')`, and that is an install artifact, not a docs defect. Keep the last sentence — "do not claim the site builds when you have not seen it build" is the valuable half and survives unchanged.

.claude/agents/docs-style-reviewer.md
```diff
@@ -0,0 +1,73 @@
...
+  sidebar is autogenerated from the directory tree, so a page without
+  `sidebar_position` lands in an arbitrary slot.
+- **New directory without `_category_.json`** (needs `label`, `position`, and a
+  `link` to its index doc).
+- **Unclosed code fences** or a fence whose language tag is missing where the site
+  otherwise uses one.
+
+Where you can, verify rather than eyeball: grep the link targets, and if the site
+builds in this environment, `npm --prefix docs run build` is the ground truth.
+Note that on Node 24 with Docusaurus 3.6.3 the build currently fails SSG for every
```
CoderCoco: **Same false claim as `SKILL.md:110`, and it lands harder here.**

This is the agent that owns build verification, and its instruction is to treat an identical failure on clean `main` as pre-existing. The build is green at this head (`npm ci && npm run build` in `docs/`, Node v24.18.0, Docusaurus 3.6.3 → `[SUCCESS]`), so the premise is wrong.

Concrete failure: this agent is dispatched on a docs change that introduces a genuine MDX parse error. It runs the build, sees a throw, matches it against "currently fails SSG for every page for reasons unrelated to any docs edit," and reports **no build risks** — which the skill's step 4 treats as the not-optional gate. The one finding class this agent exists to catch is exactly the class this paragraph teaches it to suppress.

Suggested fix: drop the 3.6.3-specific paragraph and keep only "`npm --prefix docs run build` is the ground truth." If a self-check against install state is still wanted, phrase it as a precondition (`npm ci` in `docs/` first) rather than as a recorded failure signature that an agent can pattern-match a real error onto.

CLAUDE.md
```diff
@@ -1,282 +1,181 @@
...
+- **No `as unknown as T` casts in tests.** Prefer `vi.mocked(fn)` for mocked modules and
+  `Partial<T>` + a single `as T` for service-shaped stubs.
+- **No raw `process.env` in business logic.** Wrap env access behind a service method so
+  tests stub it with `vi.spyOn` instead of mutating `process.env`.
+
+Three complementary test tiers:
+
+| Tier | Command | What runs |
+|------|---------|-----------|
+| Unit / component | `npm run app:test` | Vitest. Node environment for server logic; `jsdom` for `@hyveon/web` component and routed-page specs (co-located `Foo.test.tsx`, mounted through `renderPage()`). AWS SDK mocked via `aws-sdk-client-mock`. |
```
CoderCoco: **`renderPage()` is now named here but defined nowhere, and the conventions around it were deleted rather than moved.**

The PR body says the `__test.mock` seam "was the only content in `CLAUDE.md` with no home in `docs/`." Grepping `docs/docs/**` at this head:

```
renderPage             (no hits)
vitest.setup           (no hits)
environmentMatchGlobs  (no hits)
test-utils             (no hits)
```

So the "React component unit tests" and "Routed page tests" blocks — nine bullets — went from `CLAUDE.md` to nowhere. The load-bearing ones: co-locate `Foo.test.tsx` next to the component rather than in `__tests__`; `vitest.setup.ts` auto-loads jest-dom matchers and wires `afterEach(cleanup)` because we run `globals: false`; `environmentMatchGlobs` flips `packages/web/**` to jsdom; `vi.stubGlobal('EventSource', …)` for SSE components; no snapshots; and **stub at minimum `api.status` and `api.costsEstimate` or the test hangs**.

That last one is a live trap, not folklore — `GameStatusProvider` calls `api.costsEstimate()` unconditionally at `app/packages/web/src/polling/game-status-provider.component.tsx:72`, above every page mounted through `renderPage()`.

Concrete failure: an agent adds `TerraformHistoryPage.test.tsx`, mocks `../api.js` with only the methods the page itself calls, and the test hangs waiting on the polling registry. This row points at `renderPage()`, so the agent looks it up — and finds no page describing it, because the helper is at `src/test-utils/render-page.utils.tsx` (its own TSDoc covers the provider stack but not the required stubs) and nothing in `docs/` mentions it.

Suggested fix: move the two blocks into `docs/docs/components/integration-tests.md` alongside the tier-1 seam this PR already relocates there, and point this row at it — the same treatment, applied to the content that also qualified. Worth noting the old path was already wrong (`test-utils/renderPage.tsx`; the file is `render-page.utils.tsx`), so whatever lands should carry the corrected path.

.claude/skills/write-docs/SKILL.md
```diff
@@ -0,0 +1,148 @@
...
+## Gotchas
+
+- `.md` files are parsed as MDX. A bare `{` or `<` outside a code fence is a parse
+  error — wrap identifiers like `{game}-server` or `Partial<T>` in backticks.
+- Don't duplicate `openspec/specs/**` requirement wording into docs. Link to the
+  behaviour and explain it; two copies drift and the spec is the one that wins.
+- Don't restate what `CLAUDE.md` covers (workflow rules, PR conventions). The
+  arrow points the other way: `CLAUDE.md` points at docs.
+- Adding a new component page means also adding it to
+  `docs/docs/components/index.md`, and to the routing table in `CLAUDE.md`.
```
CoderCoco, line 154: **Cheap, and mostly a consistency point:** this rule, plus `docs-coverage-auditor`'s "a new or relocated doc that agents should route to belongs in the routing table in `CLAUDE.md`," is not satisfied by this PR itself. The rewritten routing table lists `.claude/commands/pr.md` and `.github/copilot-instructions.md` but neither this skill nor the four new agents, and the new "Before opening a PR" section mandates a docs update without naming the workflow built to produce one.

Concrete failure: an agent about to open a behaviour-changing PR reads `CLAUDE.md` top to bottom, reaches "update every page the change touches," and hand-edits the pages — never learning the writer/evaluator loop exists, since discovery depends entirely on the skill description matching the phrasing of the moment.

Suggested fix: one row in the routing table — writing or updating `docs/docs/**` → `/write-docs` — and a pointer from the docs bullet in "Before opening a PR".

CoderCoco (commented):
## Headline: the "known docs build breakage" is not real, and it is baked into two agent-facing files

I ran the build at this PR's head, on this machine, after a clean install:

```
cd <worktree at f3a10f6>/docs && npm ci && npm run build
...
[SUCCESS] Generated static files in "build".
```

Node `v24.18.0`, `@docusaurus/core@3.6.3` — the exact combination the skill and `docs-style-reviewer` record as broken. It builds, including the section this PR appends to `integration-tests.md`.

I also reproduced the failure you saw, and it has a different cause. In the main checkout, `docs/node_modules` is at 3.6.3 while `docs/package.json` (post-#352) pins 3.10.2. Building against that mismatched tree fails SSG for every page with `Cannot read properties of undefined (reading 'id')` at `DocItem` — an incremental-install artifact, not a Node-24 or Docusaurus-3.6.3 defect. PR #352's own test plan makes the same point explicitly: "3.6.3 was not broken — it builds fine under Node 24", verified with a clean `npm ci`.

That matters more than a stale sentence usually would, because of what this PR is for. The skill's stated thesis is that a confidently wrong docs page is worse than a missing one, and its single verification step is `npm --prefix docs run build`. Recording that step as "currently fails for every page, pre-existing" tells every future agent to skip the one check the workflow exists to enforce.

Second, the PR body's claim that the `__test.mock` seam "was the only content in `CLAUDE.md` with no home in `docs/`" doesn't hold — the React component and routed-page test conventions were also homeless and were dropped rather than moved.

## Findings, ranked

| # | Where | Finding | Verdict |
|---|-------|---------|---------|
| 1 | `SKILL.md:110` | "Known breakage" is false; disproven by a clean build at this head | Fix before merge |
| 2 | `docs-style-reviewer.md:33` | Same false claim, in the agent whose job is to run that build | Fix before merge |
| 3 | `CLAUDE.md:108` | React/routed-page test conventions deleted with no home in `docs/`; `renderPage()` now named but defined nowhere | Fix before merge |
| 4 | `SKILL.md:148` | The skill requires new routable docs to reach `CLAUDE.md`'s routing table; this PR's own skill and four agents aren't there | Cheap |

## What's good

The base-ref threading added in f3a10f6 is the right fix and is stated in the place it can actually be acted on — "an evaluator that guesses the base audits the wrong range and reports clean" is the correct failure mode to name. Splitting `docs-style-reviewer` into build-breaking vs. advisory, with the explicit "a reviewer who mixes them gets the mandatory ones ignored," is the kind of rationale these agent files usually lack. The four repo-specific gotchas baked into the agents (`onBrokenLinks: throw`, root-relative links with no `/docs` prefix, MDX parsing of `.md`, generated screenshots) are all correct — I verified each against `docs/docusaurus.config.ts` and the pages themselves: no `](/docs/…)` links exist anywhere, and every page under `docs/docs/**` does carry both `title` and `sidebar_position`, so `docs-style-reviewer`'s frontmatter rule won't fire false positives. The stale content the rewrite corrects is genuinely stale — five Lambdas, the missing `cloud-aws`/`desktop-preload` workspaces, and the `authGate`/`authedPage` fixtures all check out against the tree.

## Verified vs. not

**Verified, by running it:** `npm ci && npm run build` in `docs/` at this head (green); the same build against the mismatched `node_modules` in the main checkout (fails, as described above).

**Verified, by reading the tree at this head:** every npm script in the rewritten command block exists in `package.json` / `app/package.json`, and `app:build`'s stated order matches; six workspaces plus five Lambdas; `openspec/specs/in-task-tls-termination` and `desktop-only-operator-surface` exist; every page path in the routing table and in both owner-mapping tables exists; `renderPage` is exported from `app/packages/web/src/test-utils/render-page.utils.tsx` and `GameStatusProvider` does call `api.costsEstimate()` (`game-status-provider.component.tsx:72`), so the deleted warning described a real trap; `claude-md-management` is installed, so that routing pointer resolves.

**Not verified:** I did not run `app:lint`, `app:typecheck`, or any test suite — this diff is docs and agent definitions only, and I'm taking the PR body's "suites unaffected" at face value rather than re-running them. I did not exercise the skill or the four agents end to end. I did not read every `docs/docs/**` page in full to hunt for further content the `CLAUDE.md` rewrite dropped; finding 3 comes from grepping `docs/docs/**` for the specific identifiers the deleted sections named.

.claude/skills/write-docs/SKILL.md
CoderCoco: Fixed in 5545354. You are right, and I verified it rather than taking it on trust — `npm ci && npm run build` in `docs/` at this head, Node v24.18.0, `@docusaurus/core@3.6.3`:

```
[SUCCESS] Generated static files in "build".
```

So the note was false. One correction to the diagnosis: in the tree where I first hit the failure, `docs/package.json` and the installed `@docusaurus/core` both read 3.6.3, so a version mismatch was not visible at that level — I did not isolate which package was stale, only that a clean `npm ci` at this head builds green. The install-artifact conclusion holds; the specific mechanism I have not proven.

The paragraph is now the install precondition (`npm ci` first, with the `Cannot read properties of undefined (reading 'id')` signature named as an install artifact), and I kept the half you flagged as valuable plus a sharper version of it: treat a failure as yours until proven otherwise, because reporting a real error as pre-existing is the worst available outcome when `onBrokenLinks` is `throw`.

Also re-ran the build after appending the new sections — still `[SUCCESS]`.

.claude/agents/docs-style-reviewer.md
CoderCoco: Fixed in 5545354 — same root cause as the `SKILL.md:110` thread, and you are right that it lands harder here.

Dropped the 3.6.3 paragraph. The agent now runs `npm --prefix docs ci` first, and the only failure it may attribute to the environment is the stale-install signature, and only if a clean install still reproduces it. Everything else is a finding against the change until proven otherwise, with the reason stated: reporting "no build risks" because a real error resembled a known-bad signature defeats the review.

CLAUDE.md
CoderCoco: Fixed in 5545354. The claim in the PR body was wrong — those two blocks were homeless too, and I dropped them instead of moving them.

Both now live in `docs/docs/components/integration-tests.md` under "Related: unit-tier React component and routed-page specs", next to the tier-1 seam, and the `CLAUDE.md` row plus a new pointer below the Playwright bullets both route there. The `api.costsEstimate` trap is called out explicitly with the line reference you gave.

Verifying rather than moving the text verbatim turned up three more staleness bugs in the deleted blocks, so the relocated version differs from what was there:

- `environmentMatchGlobs` was removed in Vitest 4; the node/jsdom split is now two projects (`node`, `web`) inheriting via `extends: true` (`app/vitest.config.ts:44-84`).
- The helper path was wrong in two ways, as you noted — it is `src/test-utils/render-page.utils.tsx`.
- `vi.stubGlobal('EventSource', …)` has no remaining call sites; streaming mocks go through `toStreamHandleMock()` (`src/test-utils/stream-handle.test-utils.ts`), which supplies the `cancel()` a bare async generator lacks.

Docs build re-run after the append: `[SUCCESS]`.

.claude/skills/write-docs/SKILL.md
CoderCoco: Fixed in 5545354. Fair — the PR broke its own rule.

Added a routing-table row (writing or updating anything under `docs/docs/**` → the `write-docs` skill, naming `docs-writer` and the three `docs-*` evaluators), and extended the docs bullet in "Before opening a PR" to point at the skill rather than leaving discovery to whether the skill description happens to match the phrasing of the moment. Also added `docs/docs/app/*` to that bullet's page list, which was missing.

